PR #17, opened 2023-02-17
Introduce changes recommended in the kernel mailing list

teackot:
- use sysfs_emit() instead of sprintf() to prevent buffer overflow
- remove unnecessary ACPI check
- make i and len in ec_read_seq() the same type
- better error handling
- use pr_fmt for convenience


teackot:
When this gets merged, I'll also send it to LKML

BeardOverflow:
I reviewed each commit and looks good.
Thanks for your constant effort!

PR #22, opened 2023-02-19
New configurations

teackot:


teackot:
Pretty much ready to be merged. I'll send other configs in the next PR when they get ready

Issue #27, opened 2023-03-02
Port to Linux 6.2


teackot:
We need to port the dirver to Linux 6.2 as battery hook API has changed. It is just a matter of adding `struct acpi_battery_hook *hook` argument to `msi_battery_add` and `msi_battery_remove` functions.

We also need to provide a patch for distros that still use previous versions. Maybe make patching automatic.

It is already fixed in the upstream kernel patch, so we only need to catch up with it here.

PR #29, opened 2023-03-04
Port to Linux 6.2

teackot:
#27

Patch for Linux < 6.2 is provided


BeardOverflow:
Hello @teackot 
In order to support older and newer kernels, your patch would be more generic, using the `KERNEL_VERSION` macro. For example:
```
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,2,0)
static int msi_battery_add(struct power_supply *battery,
			   struct acpi_battery_hook *hook)
#else
static int msi_battery_add(struct power_supply *battery)
#endif
```
However, remember that in the official Linux' staging tree should not be typed any LINUX_VERSION_CODE. So, my proposal is:  just keep the patch, but using KERNEL_VERSION macro and renaming it to a generic name like older-kernel.

teackot:
Hi! I updated and renamed the patch considering your remarks

BeardOverflow:
Great, now your PR is totally ready.
Thanks a lot for providing this patch, it will be useful for older linux distros.

PR #34, opened 2023-03-14
Move list of tested laptops to up and remove deprecated message about constants.h.Add MSI summit e14 evo to list.

giga-bytes-dev:


BeardOverflow:
Why does not I reply this? Sorry for my late answer @giga-bytes-dev
In order to improve your PR, could you recollect all supported laptop models and adding to the README.md file? If you cannot do it, I just merge your PR as-is.

teackot:
Closing since the list is outdated (and will eventually be moved to a different file / wiki) and the message has been changed

PR #40, opened 2023-05-17
Add MSI Prestige 15 and MSI Modern 15 support

LucaPatarca:


BeardOverflow:
Thanks for your first-time contribution.

serfreeman1337:
Can we have support for updated bios as well?
` msi_ec: Firmware version is not supported: '1552EMS1.120'`

Issue #48, opened 2023-06-03
1552EMS1 duplicated


glpnk:
check ```ALLOWED_FW_11``` and ```ALLOWED_FW_2```

In FW_2 version ```1552EMS1.118```
In FW_11 version ```1552EMS1.115```

Also I think that usage of one config for two+ models is bad idea for now

serfreeman1337:
Also module refuses to load on Modern A11M with updated bios:
`msi_ec: Firmware version is not supported: '1552EMS1.120'`

teackot:
The major version of the EC FW is the same. The only difference is the minor version (115 and 118), so they should probably have the same addresses. We can read the official MSI changelogs to see if they made any major changes

teackot:
> We can read the official MSI changelogs to see if they made any major changes

Can't find them though

teackot:
CONF2 was taken from @ThePBone's [msi-ec-modern](https://github.com/ThePBone/msi-ec-modern/) repo.

CONF11 was added by @LucaPatarca 

@LucaPatarca Can you please show us an EC dump from the debug mode? (`sudo make reload-debug` and then `cat /sys/devices/platform/msi-ec/debug/ec_dump`)

glpnk:
Well, as #40 and [`msi-ec-modern`](https://github.com/ThePBone/msi-ec-modern/) fork

+ `1552EMS1` is WMI2 device, because intel 11 gen
+ fan mode address is `0xd4`, which is right
+ preset memory table mentions `0xd2` as shift, `0xd3` as keyboard, `0xd4` as fan

Only incorrect value in `conf_2` was shift mode addres, which should be `0xd2`

cvlc12:
Hi,
I just noticed I still have this inconsistency on my Modern 15 A11M:
```
kernel: DMI: Micro-Star International Co., Ltd. Modern 15 A11M/MS-1552, BIOS E1552IMS.121 04/13/2022
kernel: msi_ec: Firmware version is not supported: '1552EMS1.120'
```
Using `6.15.8-arch1-1`

glpnk:
You was wrong, also your device is supported

cvlc12:
What do you mean?

glpnk:
read #382 and try to fix your installation issues and never ever use `make install`

PR #36, opened 2023-04-15
Add memory map comparison 

glpnk:
Add text representation of address space for each data structures + some addresses from 14JKEMS1 EC analysis

PR #49, opened 2023-06-03
Supported devices list

glpnk:
Part of project documentation

Plans for next PRs:
+ rewrite memory map
+ add RWEverything guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,10 @@ Also, and until future enhancements, no DMI data is used to identify your laptop
 1. Install any AUR helper ([yay](https://github.com/Jguer/yay) for example)
 2. Run `yay -S msi-ec-git`
 
+## Supported devices and contribution
+
+[List of supported devices](docs/TOH.md)
+
 ## Usage
 
 This driver exports a few files in its own platform device, msi-ec, and is available to userspace under:
@@ -178,10 +182,3 @@ Led subsystem allows us to control the leds on the laptop including the keyboard
     - 1: On
     - 2: Half
     - 3: Full
-
-
-## List of tested laptops:
-
-- MSI GF75 Thin 9SC (17F2EMS1.106)
-- MSI Prestige 15 A11SCX (16S6EMS1.111)
-- MSI Modern 15 A11M (1552EMS1.115)
--- a/docs/TOH.md
+++ b/docs/TOH.md
@@ -1,0 +1,34 @@
+# Table of Hardware
+
+Models with same EC name on support page was combined in single generation
+
+| Model                              | Class    | EC name            | Extras                                           | Link                                                                                                                       |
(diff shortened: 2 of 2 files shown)


teackot:
How about listing the supported features, something like:

| Model                              | Class    | EC name            | Extras                                           | Link                                                                                                                       | Battery thresholds | webcam | ... |
|------------------------------------|----------|--------------------|--------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|:---:|:---:|:---:|
Cool Laptop 15 | Business | 12345678 | | | + | - | ... |

glpnk:
@teackot my last thougths was about using `Discussions` for supported models and other documentation

teackot:
Like using discussions as docs pages? I think this way docs will be lost among questions and firmware requests

We can also use github's wiki feature for this

teackot:
Though the issue with the wiki is that only I and BeardOverflow will be able to edit it. But it would work nice for documenting different features the driver exposes in sysfs. That is, "Usage" section of the current readme.md

glpnk:
This is my second attempt to create "low level" documentation. Also how to implement usable memory map table from #36.

I think mixing code with docs inside same repo is bad, because at least it messing with commit history.

glpnk:
@teackot How do you think about creating repo for wiki repo content? Like you do for issue template

teackot:
> What do you think about creating repo for wiki repo content? Like you do for issue template

I only created issue template repo as a demonstration, it is already merged into this repo. I think creating separate repos for stuff will cause too much fragmentation for a small driver project

teackot:
Oh, I just found out that github wiki can be edited by contributors and is in fact a separate git repo that you can clone and edit locally, it just needs to be enabled in the repo settings

teackot:
Yes

We can make a single contribution guide with a Windows section, a Linux section, and a section about creating a new configuration in the code

glpnk:
@teackot Okay. Not related to this theme but what you think about #48 and similar cases and single `FW` record for two+ different models? Or it pointless in case of possible structural redesign?

teackot:
I will reply in #48

teackot:
@glpnk BeardOverflow gave contributors access to the wiki, see if you can edit it

rottenpants466:
Maybe do something similar to this? 

https://github.com/dmitry-s93/MControlCenter/blob/main/docs/tested_devices.md

teackot:
@rottenpants466 

I think it's a good idea to do a similar table. We already have a list of laptops in README and we can just put them in a separate file with a simple list of features for each machine.

I also think we should put it in the /docs directory of the repo instead of the wiki. Wiki is nice, but it is harder to modify and keep track of.

teackot:
See: #116

glpnk:
Deprecated, feel free to reuse

PR #50, opened 2023-06-10
Raider GE78HX SmartTouchpad offsets

Myp3a:
Offsets for my Raider GE77HX. Found by RWEverything, all of them works in Windows and Linux. However, FN key swap returns opposite values.

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -2138,6 +2138,85 @@ static struct msi_ec_conf CONF25 __initdata = {
 	},
 };
 
+static const char *ALLOWED_FW_26[] __initconst = {
+	"17S2IMS1.113", // Raider GE78 HX Smart Touchpad 13V
+	NULL
+};
+
+static struct msi_ec_conf CONF26 __initdata = {
+	.allowed_fw = ALLOWED_FW_26, // WMI2 based
+	.charge_control = {
+		.address      = 0xd7,
+		.offset_start = 0x8a,
+		.offset_end   = 0x80,
+		.range_min    = 0x8a,
+		.range_max    = 0xe4,
+	},
+	.webcam = {
+		.address       = 0x2e,
+		.block_address = 0x2f,
+		.bit           = 1,
+	},
+	.fn_win_swap = {
+		.address = 0xe8,
+		.bit     = 4,
+		.invert  = true,
+	},
+	.cooler_boost = {
+		.address = 0x98,
+		.bit     = 7,
+	},
+	.shift_mode = {
+		.address = 0xd2,
+		.modes = {
+			{ SM_ECO_NAME,     0xc2 },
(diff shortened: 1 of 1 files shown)


teackot:
Looks good to me

> However, FN key swap returns opposite values

That's a known issue. I think we should just make it an on/off setting, i.e. swap/not swap

teackot:
> However, FN key swap returns opposite values

What does it say when those keys are swapped?

teackot:
@Myp3a can you please attach an EC dump, so we can verify the WMI 'version' and also just for the record:

```bash
sudo make load-debug
cat /sys/devices/platform/msi-ec/debug/ec_dump
```

> However, FN key swap returns opposite values

Already fixed

glpnk:
@Myp3a Why you're keeping old commits after reverting it in second commit?

Please add full name of your device which may look like `Raider GE78 HX Smart Touchpad 13V` (just first link from Google)

Myp3a:
Unfortunately, my GE77HX broke, so I'm on GE78HX SmartTouchpad now. They are pretty similar, however. Both use the Steelseries keyboard, so no backlight control from the EC.  
I've updated the PR accordingly:
 - FW version is changed to GE78HX
 - Added FN key invert
 - Added mute/mic leds

Here is the debug info:
FW version: 17S2IMS1.113
FW release date: 2023/11/27 16:21:55
EC dump: [Raider_GE78HX_STP_Dump.txt](https://github.com/user-attachments/files/16216507/Raider_GE78HX_STP_Dump.txt)

glpnk:
Ok, thanks for explanation. But please use full name from MSI site, which include CPU generation https://www.msi.com/Laptop/Raider-GE78-HX-Smart-Touchpad-13VX/Specification

glpnk:
Thank you

Myp3a:
Yeah, I've messed up a bit with merge conflicts. Should be resolved now, thank you for the note

glpnk:
Another laptop with keyboard size touchpad #120.

@Myp3a Do you have any problems with touchpad on Linux? Like the ones described in [this comment](https://github.com/BeardOverflow/msi-ec/pull/120#issuecomment-2227100074)?

Myp3a:
I'm using headless Linux, so can't really tell. However, from my experience, there's no pressure sensor in the touchpad, only the click mechanism. Also, all touchpad keys are controlled from software, not the EC - so touchpad keys support is out of the scope of this project.
Btw, I've made a small tool to remap the keys - they are used together with raw input messages. Probably will look into porting it to Linux in the future.

glpnk:
Nice. Is it using WMI event subscription? Or touchpad driver

Myp3a:
Not really. Touchpad sends raw keyboard messages, which contains the ID of pressed key - and the software part recognizes it and does something

teackot:
Merged! 🎉

PR #58, opened 2023-07-18
Remove BIOS versions from allowed firmware list

rottenpants466:
Added a new firmware version: E1592IMS.10F

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -298,6 +298,7 @@ static const char *ALLOWED_FW_3[] __initconst = {
 	"1592EMS1.111",
 	"E1592IMS.10C",
 	"E1592IMS.10D",
+	"E1592IMS.10F",
 	NULL
 };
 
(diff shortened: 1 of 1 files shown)


teackot:
Does this PR make the driver work on new firmware? It seems that the versions that start with E are BIOS FW versions, not EC FW

rottenpants466:
> 

It works with this PR prior to kernel 6.4.

This is the change log from msi:

New BIOS     : 	E1592IMS.10F 
ROM CheckSum :  0x1B27 
Release Date :  2024/7/14

;--------------------------- Description ------------------------------------;

1.Enhanced security protection against unauthorized BIOS updates.

teackot:
Okay then, looks good to me

teackot:
Okay, so now that we know that `E1592IMS.10C/D/F` is actually BIOS and not EC, can you please remove those three from the allowed fw array? You can even do it in this PR, just rename it after

teackot:
Yep, everything's correct, thank you too!

PR #59, opened 2023-07-21
msi-ec : add support gp66-11ug

perseusdg:
This pull request adds support for msi gp66-11ug, I am guessing that this commit should work with other gp66-11u variants and gp76-11u variants.

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -995,6 +995,88 @@ static struct msi_ec_conf CONF11 __initdata = {
     },
 };
 
+static const char *ALLOWED_FW_12[] __initconst = { 
+    "1543EMS1.113", 
+	NULL 
+};
+
+static struct msi_ec_conf CONF12 __initdata = {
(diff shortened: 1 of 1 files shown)


teackot:
Hi! Have you tested all of the features? Also can you please attach an EC dump (`/sys/devices/platform/msi-ec/debug/ec_dump` with debug mode on) here, so we can use it as a reference in the future

perseusdg:
I did test out most of them , the only thing i couldn't verify properly was the gpu and cpu fan speed..I will attach the ec dump in a couple of days

teackot:
I think that since you've tested it out, I can merge it. Some dumps woud still be appreciated though

Thank you!

teackot:
Actually, we have a dump in #95

NamitBhutani:
Hi, I have a MSI GP76 11UG, and advanced fan speed control doesn't seem to work on nixos. Cooler boost works and disabling the webcam works tho.

PR #67, opened 2023-08-22
Fix negative start threshold

### LINKED ISSUE #56: Charge control broken on MSI Alpha 15 B5EEK
I have an msi alpha B5EEK-070 with bios version: E158LAMS.10B and linux 6.4.
Charge control doesn't seem to work, both charge_control_start_threshold and charge_control_end_threshold report strange values and when I tried to change charge_control_end_threshold the fan started to spin very fast.
```
cat /sys/class/power_supply/BAT1/charge_control_start_threshold
-10
cat /sys/class/power_supply/BAT1/charge_control_end_threshold
0
```

teackot:
Fixes #56

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1164,6 +1164,11 @@ static ssize_t charge_control_threshold_show(u8 offset, struct device *device,
 	if (result < 0)
 		return result;
 
+	// thresholds are unknown
+	if (rdata == 0x80) {
+		return sysfs_emit(buf, "0\n");
+	}
+
 	return sysfs_emit(buf, "%i\n", rdata - offset);
 }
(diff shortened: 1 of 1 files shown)


rottenpants466:
Maybe the same should be done to the end threshold values. In my case there was a start -138 and a end -128: https://bugs.kde.org/show_bug.cgi?id=473281

That's why the option to change the thresholds were not appearing in KDE power settings :)

cheers

teackot:
As we already found out, your problem was because of the wrong charge control address. So if the value is anything other than 10%-100% or 0x80 then it is better for it to not show up in the settings bc the address is probably wrong

PR #84, opened 2023-10-09
Avoid dynamic memory allocation

teackot:
Use static arrays when filtering attributes to avoid potential memory issues


teackot:
Solved by #224

Issue #105, opened 2024-01-27
MSI Titan 18 HX A14V - Firmware version is not supported 


nalim:
### Laptop model

MSI Titan 18 HX A14V

### EC firmware version

1822EMS1.105

### EC memory dump
```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 00 27 1b 5b
0x3_ | 03 01 00 0d 00 00 51 81 6a 18 60 3b 71 02 c0 00
0x4_ | 35 0c 58 00 b9 18 00 00 a6 15 2d 41 d9 0b f8 43
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 34 00 3a 40 46 4c 52 58
0x7_ | 64 19 00 19 23 2c 3a 46 4d 4b 09 03 03 03 03 03
0x8_ | 00 00 34 3a 40 46 4c 52 64 00 00 19 23 2c 3a 46
0x9_ | 4d 5f 07 03 03 03 03 03 02 00 6e 02 00 66 40 00
0xa_ | 31 38 32 32 45 4d 53 31 2e 31 30 35 31 31 32 39
0xb_ | 32 30 32 33 31 31 3a 31 33 3a 31 38 00 00 00 08
0xc_ | 00 00 07 22 00 00 19 00 01 1a 00 00 00 00 00 00
0xd_ | 00 00 c1 84 0d 00 05 bc 00 01 00 00 00 05 00 00
0xe_ | e2 00 00 b9 18 00 00 c1 01 00 00 00 00 92 00 00
0xf_ | 00 00 70 00 2a 7f 05 28 7f 05 00 00 00 01 00 00
```
### GPU

Nvidia

### Is your keyboard RGB?

Yes (multi color RGB)

### Additional context

https://www.msi.com/Laptop/Titan-18-HX-A14VX/Specification

Makintos:
New BIOS and EC firmware version released two days ago:
```
[...]

PR #120, opened 2024-05-16
Titan 18 HX A14V [1822EMS1]

glpnk:
#105

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1498,8 +1498,8 @@ static struct msi_ec_conf CONF18 __initdata = {
 	// 	.bit          = 5,
 	// }, // Like Katana 17 B11UCX
 	.webcam = {
-		.address       = 0x2e,
-		.block_address = 0x2f,
+		.address       = MSI_EC_ADDR_UNSUPP,
+		.block_address = MSI_EC_ADDR_UNSUPP,
 		.bit           = 1,
 	},
 	.fn_win_swap = {
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1480,6 +1480,7 @@ static const char *ALLOWED_FW_18[] __initconst = {
 	"1822EMS1.105", // Titan 18 HX A14V
 	"1822EMS1.109", // WMI 2.8
 	"1822EMS1.111",
+	"1822EMS1.112",
 	NULL
 };
 
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1476,6 +1476,90 @@ static struct msi_ec_conf CONF17 __initdata = {
 	},
 };
 
+static const char *ALLOWED_FW_18[] __initconst = {
+	"1822EMS1.105", // Titan 18 HX A14V
+	"1822EMS1.109", // WMI 2.8
+	"1822EMS1.111",
+	NULL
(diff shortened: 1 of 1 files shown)


sempervictus:
Probably a good idea to introduce a "MAYBRICK" danger tag or something like that to help bots which build PRs avoid by tag.

glpnk:
I think someone tested this version, but I need to recheck something. Basically, many laptops have similar configs, but some features are unavailable

sempervictus:
@glpnk - anything i can do to help this PR along?

glpnk:
Looks like ready, but require checking ability to set keyboard backlight, because some SteelSeries keyboard don't use EC, but MSI program writing dummy values

Thanks for waiting, @makintos @nalim @sempervictus 
Thanks to @makintos for research

sempervictus:
Seems to work for battery threshold anyway:
```
# /etc/systemd/system/battery-charge-save-threshold.service
[Unit]
Description=Set the battery charge threshold to max lifetime
After=multi-user.target
StartLimitBurst=0
#Conflicts=battery-charge-full-threshold,battery-charge-balanced-threshold

[Service]
Type=oneshot
Restart=on-failure
ExecStart=/bin/bash -c 'echo min > /sys/devices/platform/msi-ec/battery_mode && echo "Battery charge threshold $(cat /sys/devices/platform/msi-ec/battery_mode)"'

[Install]
WantedBy=multi-user.target
```
Hasn't blown up my laptop or thrown any errors in compilation using GCC14.1 with pretty much every Grsec/PaX plugin enabled.

glpnk:
@sempervictus 
Have you tested ability to change keyboard brightness from linux?

sempervictus:
Keyboard brightness works, touchpad `fn` key does not. Touchpad is actually kind of weird - using `libinput` with X, i am finding that the "physical click" pressure changes becoming _very_ sensitive to accidental contact when typing and that the "disable touchpad while typing" function in Cinnamon's touchpad settings does not work. Otherwise, super-battery mode, charge adjustment thresholds, keyboard brightness, microphone, fan speed, and volume all work as expected

glpnk:
@sempervictus 

> touchpad fn key does not

What? How this look like? And how should work?

How actually super battery works? Is it reduces CPU/GPU speed? Because I'm not have it.

image
Is this really keyboard sized touchpad?

sempervictus:
Touchpad is about size of space bar integral to the surface. The whole area has mechanical click and it appears to be pressure adjustable but that adjustment sort of swims randomly.
Superbat slows all the chips, kills discrete GPU AFAIK (no idea, can't use nvidia drivers with any form of CFI as they perform function hijacking), and limits fans accordingly. Seems to get about an hour of added life

glpnk:
Thanks, you probably need to ask community of your touchpad manufacturer, if it is really possible

teackot:
TODO: docs

teackot:
@glpnk we still need to implement those unusual kbd backlight intensity values, right?

glpnk:
@teackot we just ignore keyboard brightness for RGB and per key RGB keyboards, value in EC just cycles and MSI app receive key press events

Issue #99, opened 2023-12-18
MSI Alpha 17 C7VG - Firmware version is not supported


Roh8i:
### Laptop model

`Alpha 17 C7VG`

### EC firmware version

`17KKIMS1.108`

### EC memory dump

```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 00 00 0b 52
0x3_ | 03 09 00 0d 01 00 50 81 6a 18 60 3b 71 02 e0 00
0x4_ | 00 00 63 00 60 18 00 00 fb 17 00 43 9a 0b 00 00
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 37 00 32 39 40 47 4e 55
0x7_ | 6c 28 00 28 2d 32 39 41 50 69 04 04 04 04 04 09
0x8_ | 00 00 37 3c 41 46 4b 50 62 00 00 30 37 3e 44 4d
0x9_ | 55 64 04 04 04 04 04 04 06 00 00 06 00 00 00 00
0xa_ | 31 37 4b 4b 49 4d 53 31 2e 31 30 38 30 39 32 30
0xb_ | 32 30 32 33 31 36 3a 30 39 3a 32 35 00 00 00 00
0xc_ | 00 00 06 22 00 00 00 00 00 cd 00 cd 00 00 00 00
0xd_ | 00 00 c1 00 0d 00 05 80 00 00 00 00 00 00 00 00
0xe_ | e2 00 00 5d 18 00 00 40 00 00 00 00 00 9a 00 00
0xf_ | 40 00 70 00 26 80 23 60 20 90 00 00 00 01 00 00

```

### EC firmware version

`17KKIMS1.109`

### EC memory dump

```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[...]

Issue #96, opened 2023-11-30
Bravo 15 C7V firmware version is not supported


je-miralles:
### Laptop model

Bravo 15 C7V

### EC firmware version

158NIMS1.109

### EC memory dump

```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 08 00 0b 5a
0x3_ | 03 01 00 05 00 00 50 81 d2 11 88 2c c8 01 c0 00
0x4_ | f8 11 50 00 b3 11 00 00 1a 0e e0 2f b4 0b fa 32
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 3a 00 3c 41 46 4e 55 5a
0x7_ | 64 00 00 2d 36 41 46 4b 50 64 0a 03 03 03 03 03
0x8_ | 39 00 3c 41 46 4b 50 55 64 00 00 2d 36 41 46 4b
0x9_ | 50 64 0a 03 03 03 02 03 06 00 00 06 00 00 00 00
0xa_ | 31 35 38 4e 49 4d 53 31 2e 31 30 39 30 39 31 39
0xb_ | 32 30 32 33 31 30 3a 32 37 3a 31 37 00 00 00 00
0xc_ | 00 00 06 22 00 00 00 00 00 00 00 00 00 00 00 00
0xd_ | 00 00 c1 81 0d 00 05 d0 00 00 00 00 00 00 00 00
0xe_ | e2 00 00 b3 11 00 00 40 00 00 00 00 00 9a 00 00
0xf_ | 00 00 70 00 37 20 00 00 00 00 00 00 00 00 00 00
```

### GPU

Nvidia

### Is your keyboard RGB?

No (single color)

### Additional context

Regarding keyboard RGB: I have only seen a single colour, but unsure if it's a configuration thing.

Additional ec_dumps with keyboard RGB toggles: https://gist.github.com/je-miralles/d82342fc3e5e7209832b59f4c2affcd9

je-miralles:
Quick update with my findings after staring at some EC dumps.

[...]

PR #125, opened 2024-05-30
Documentation: model naming in sources; fix: resolve Conf2-11 conflict

glpnk:
Done

Changelog:
+ add WMI/EC layout generation
+ fix naming
+ fix spaces to tabs in conf 11
+ ~~mark conf 11 as todo for fixing mixed models~~ DONE

More changes WIP: https://github.com/glpnk/msi-ec/tree/docs/more-docs/docs

---
we probably need to add option `SteelSeries` to keyboard dropdown in request template


teackot:
> what you can say about this (WIP) ...

I like the format. Though the "EC generations, WMI version" section has to be more elaborate on what we consider to be generations, and the similarities between layouts.

> we probably need to add option `SteelSeries` to keyboard dropdown in request template

Could you clarify what you mean by that?

glpnk:
You can merge this PR, because all WIP things I do in other branch. I just need to update data about latest added device before you merge this

> EC generations, WMI version

Temporary name + WIP

> Could you clarify what you mean by that?

SteelSeries appears to be USB only keyboard + KB backlights + RGB case light, and don't use EC backlight level. MSI Center writes backlight level to EC, but it seems to be not used by keyboard.
Some keyboards have FN shortcuts highlighting. But after FN-WIN swapping, highlighting is not changed. I don't know how FN-WIN swap works and if it really works.

glpnk:
I probably found a way to check WMI generation from EC dump, WMI2 module upstreamed by Wer-Wolf have function which gets EC RAM version + checks `Tiger Lake` flag. This value is `GET_EC`[1] WMI method, which corresponds to `0x37` address.

On WMI1 devices, this byte is not used, but on WMI2 it mostly equals `0x81`=`0b10000001`, so 7 bit is up, which is `Tiger lake` flag.

Tiger lake is Intel 11 generation, but this bit seems be up on 11+ gen Intel and gaming 7 gen AMD, you can check this on #105 #99 #96

teackot:
> On WMI1 devices, this byte is not used, but on WMI2 it mostly equals 0x81=0b10000001, so 7 bit is up, which is Tiger lake flag.

Great, also needs a docs mention. We might want some more additional ways to check the WMI version to check if there are any exceptions.

glpnk:
I'm thinking, how we should call device generations: WMI version or just generation. Because ACPI-WMI2 is name of MSI library, which was used in this Windows EC dumper https://github.com/timschneeb/MsiEcRamEditor

teackot:
I've seen a reference to WMI namespace versions [here](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/dn798287(v=ws.11)), not sure if it is related. I think we can go with WMI versions or WMI layouts?

glpnk:
This is [decompiled BMOF dump](https://github.com/BeardOverflow/msi-ec/discussions/98#discussioncomment-8376527), which MSI share inside their programs. Basically, WMI2/generation 2 (call it like you want) have method `Get_WMI` which returns WMI version, which equals `2.8` (for device DSDT dump of which I've used to create ImHex pattern), but only for new devices. 

So WMI2 is simple name, but confusing

glpnk:
There seems to be only 1 correct way to check EC generation / WMI version - to compare exported WMI GUIDs/UUIDs with IDs from BMOF, because some EC bytes might be changed.

Also, some bytes are mapped identically for both generations, so we can always export them

teackot:
> EC layout version may be

That's roughly what we've been calling it, it does make sense

Srry for not participating in the discussion. Right now I will focus on merging all those PRs while I'm not burnt out and while I have time

glpnk:
@teackot, ready

Also, I've fixed `conf2 - conf11` conflict https://github.com/BeardOverflow/msi-ec/issues/48#issuecomment-2179636418

glpnk:
FINALLY ready

teackot:
Yay, merging

PR #126, opened 2024-05-30
Added support for MSI GS66 12UGS

Kinue72:
### Laptop Model
MSI GS66 12UGS

### EC firmware version
16V5EMS1.107

### EC memory dump
```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 00 04 3b 4b
0x3_ | 03 09 00 0d 03 00 50 81 6a 18 60 3b 71 02 e0 00
0x4_ | 00 00 62 00 81 14 00 00 fb 13 ae 40 1c 0c 00 00
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 3d 00 37 3c 41 46 5a 5f
0x7_ | 64 41 2d 32 3c 41 50 55 64 00 08 07 05 05 05 03
0x8_ | 37 00 32 3c 46 52 5a 5d 64 2d 00 2d 3c 46 50 55
0x9_ | 64 00 03 05 05 06 03 03 02 0f 78 02 0f 78 3e 00
0xa_ | 31 36 56 35 45 4d 53 31 2e 31 30 37 31 32 31 34
0xb_ | 32 30 32 32 31 35 3a 33 30 3a 31 37 00 00 00 08
0xc_ | 00 00 07 25 00 00 00 00 00 9b 00 9d 00 c0 00 00
0xd_ | 00 00 c1 83 0d 00 05 80 00 01 00 00 00 08 00 00
0xe_ | e2 00 00 81 14 01 00 40 00 00 00 00 00 c2 00 00
0xf_ | 00 00 70 00 36 7f 05 37 41 00 03 c0 00 03 01 00
```
### GPU
Intel & Nvidia

### Is your keyboard RGB?
Per key RGB

### Additional context
Both `rt_temp_address` and `bs_fan_speed_address` are valid addresses but need tweak the `base_min` and `base_max` values. Keyboard backlighting control isn't working, it seems to be controlled by Steelseries GG. Also, FN key swap returns opposite values.

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1636,6 +1636,87 @@ static struct msi_ec_conf CONF19 __initdata = {
 	},
 };
 
+static const char *ALLOWED_FW_20[] __initconst = {
+	"16V5EMS1.107", // MSI GS66 12UGS
+	NULL
+};
+
+static struct msi_ec_conf CONF20 __initdata = {
+	.allowed_fw = ALLOWED_FW_20,
+	.charge_control = {
+		.address      = 0xd7,
+		.offset_start = 0x8a,
+		.offset_end   = 0x80,
+		.range_min    = 0x8a,
+		.range_max    = 0xe4,
+	},
+	// .usb_share  {
+	// 	.address      = 0xbf,
+	// 	.bit          = 5,
+	// },
+	.webcam = {
+		.address       = 0x2e,
+		.block_address = 0x2f,
+		.bit           = 1,
+	},
+	.fn_win_swap = {
+		.address = 0xe8,
+		.bit     = 4,
+	},
+	.cooler_boost = {
+		.address = 0x98,
+		.bit     = 7,
+	},
+	.shift_mode = {
+		.address = 0xd2,
+		.modes = {
+			{ SM_ECO_NAME,     0xc2 }, // super battery
+			{ SM_COMFORT_NAME, 0xc1 }, // balanced
+			{ SM_TURBO_NAME,   0xc4 }, // extreme
+			MSI_EC_MODE_NULL
+		},
+	},
+	.super_battery = {
+		.address = 0xeb,
+		.mask    = 0x0f,
+	},
+	.fan_mode = {
+		.address = 0xd4,
+		.modes = {
+			{ FM_AUTO_NAME,     0x0d },
+			{ FM_SILENT_NAME,   0x1d },
+			{ FM_ADVANCED_NAME, 0x8d },
+			MSI_EC_MODE_NULL
+		},
+	},
+	.cpu = {
+		.rt_temp_address       = 0x68,
+		.rt_fan_speed_address  = MSI_EC_ADDR_UNKNOWN, // 0xc9
+		.rt_fan_speed_base_min = 0x00, // ?
+		.rt_fan_speed_base_max = 0x3d, // ?
+		.bs_fan_speed_address  = MSI_EC_ADDR_UNKNOWN, // 0xcd
+		.bs_fan_speed_base_min = 0x00, // ?
+		.bs_fan_speed_base_max = 0x0f, // ?
+	},
+	.gpu = {
+		.rt_temp_address      = 0x80,
+		.rt_fan_speed_address = 0xcb,
+	},
+	.leds = {
+		.micmute_led_address = MSI_EC_ADDR_UNSUPP,
+		.mute_led_address    = MSI_EC_ADDR_UNSUPP,
+		.bit                 = 1,
+	},
+	.kbd_bl = {
+		.bl_mode_address  = MSI_EC_ADDR_UNSUPP,
+		.bl_modes         = { },
+		.max_mode         = 1,
+		.bl_state_address = MSI_EC_ADDR_UNSUPP,
+		.state_base_value = 0x80,
+		.max_state        = 3,
+	},
+};
+
 static struct msi_ec_conf *CONFIGURATIONS[] __initdata = {
 	&CONF0,
 	&CONF1,
@@ -1657,6 +1738,7 @@ static struct msi_ec_conf *CONFIGURATIONS[] __initdata = {
 	&CONF17,
 	&CONF18,
 	&CONF19,
+    	&CONF20,
 	NULL
 };
 


glpnk:
Thanks!

Do you have similar to photo keyboard layout or regular [ ctrl | fn | win | alt ]
MSI Cyborg 15 A12VF

Kinue72:
Image took from [NotebookCheck](https://www.notebookcheck.net/MSI-Stealth-GS66-12UGS-gaming-laptop-review-Powerful-enough-for-QHD.600819.0.html#8311698-1)

img

Kinue72:
After some investigation, when you hold FN, the keyboard led is the highlight key that supports the function. If swap in EC memory, it swaps the key, but the highlight is wrong. For example, my current FN is actually FN, but when I hold the left WIN key, it highlights but is still functional as a WIN key. The fn key has, but it is actual fn. When I opened MsiCenter in DnSpy, I found that this efi variable might store data for fn swap and power share `MsiDCVarData-DD96BAAF-145E-4F56-B1CF-193256298E99`

glpnk:
I don't think EFI vars is used as an interface to change settings, at least not directly. But it could be used to apply changes after reboot.

Settings being written to hardware via WMI interface, but EFI vars is like cache, I think

glpnk:
Via WMI interface on WMI2 devices, you can read/write EC RAM values. https://github.com/timschneeb/MsiEcRamEditor

But it only works for WMI2 devices, which are Intel 11+ gen and AMD 7 gen (only gaming series) CPU based devices

glpnk:
Windows has a way to log WMI requests, which look similar to SQL language SELECT queries. But it's messed. I've used this tool on my WMI1 device, but don't get anything interesting. On WMI2 should be better. https://github.com/Alois-xx/WMIWatcher

If you want to do deep dive into this check #98

Kinue72:
All offsets are copy from MsiCenter except the UEFI variable one, maybe it refresh cache or something when reboot?

glpnk:
> my current FN is actually FN, but when I hold the left WIN key, it highlights but is still functional as a WIN key 

From some laptop videos I've learned that in some devices keyboard controller and EC is different microcontrollers, but for you looks like only RGB controller is separate. So you change settings in EC/KB-controller, but RGB controller do not received that changes and highlight older layout

glpnk:
> All offsets are copy from MsiCenter except the UEFI variable one, maybe it refresh cache or something when reboot?

IDK, but on Linux I can't change EFI vars, for changing like boot order and other boot settings. This might work on Windows, but when I last time dumped these vars it was zeroed

glpnk:
On the latest kernel or soon, you can use `lm-sensors` to get a cooler RPM.

glpnk:
So your keyboard seems reporting backlight level, but when you change value in EC nothing happens?

Kinue72:
> So your keyboard seems reporting backlight level, but when you change value in EC nothing happens?

Yes, I see changed in EC memory but when write nothing happens.

glpnk:
> I found that this efi variable might store data for fn swap and power share MsiDCVarData-DD96BAAF-145E-4F56-B1CF-193256298E99

Actually, BIOS may use this variable to communicate with MSI Center about these settings to keep them in sync.

But for me, when I swapped EC-WIN from MSI Center Pro or EC, this setting just vanished from BIOS

glpnk:
Does FN key highlighting works on Linux?

Kinue72:
> Does FN key highlighting works on Linux?

It works but it desync when I swap fn and win.

glpnk:
How your keyboard is detected? As USB device? Or USB only for backlight control. But if there are any USB devices, it may use SMBUS/I2C interface

Kinue72:
> How your keyboard is detected? As USB device? Or USB only for backlight control. But if there are any USB devices, it may use SMBUS/I2C interface

IDK, Here is my `lsusb`
image

glpnk:
Someone experimented with it in `OpenRGB` on similar device https://gitlab.com/CalcProgrammer1/OpenRGB/-/issues/2642

Probably on FN-WIN swap something inside MSI Center or its backend triggers SteelSeries driver

heeen:
I added 16V5EMS1.108 to the list of supported firmwares and it loads but setting the fan mode to silent does not seem to have any effect

glpnk:
Which device you used as reference?

heeen:
I was using isw before, specifically this fork: https://github.com/YoyPa/isw/pull/245/commits/36af65bc00919b659faf301b8fcc4ed3a71eccca

do the addresses and values have any correlation to the config values in this module?

heeen:
looks like isw just uses advanced mode with a specific curve of fan speed values, am I right? I would like to replicate the same using msi-ec

glpnk:
> I was using isw before, specifically this fork: [YoyPa/isw@36af65b](https://github.com/YoyPa/isw/commit/36af65bc00919b659faf301b8fcc4ed3a71eccca)
> 
> do the addresses and values have any correlation to the config values in this module?

Yes, this module changes the same values but in different way.

> looks like isw just uses advanced mode with a specific curve of fan speed values, am I right? I would like to replicate the same using msi-ec

Yes, but this module not support fan curve changing yet. You can use https://github.com/dmitry-s93/MControlCenter

heeen:
mcontrolcenter, like isw, depends on ec_sys with write support though, right? I want to use secure boot, which is why I am moving to this module. Every time my battery runs out, the bios resets or forgets the secure boot disabled setting and I am tired of it.

heeen:
in isw the fan mode is 140, which is hex 0x8c, but here it has
```
.fan_mode = {
		.address = 0xd4,
		.modes = {
			{ FM_AUTO_NAME,     0x0d },
			{ FM_SILENT_NAME,   0x1d },
			{ FM_ADVANCED_NAME, 0x8d },
			MSI_EC_MODE_NULL
		},
	},
```

0x8d for advanced mode - is this a off by one error or am I missing something?

glpnk:
> `0x8c` or `0x8d`

Seems like this is not affecting the end result. So you can use already PRed values you mentioned.

> mcontrolcenter, like isw, depends on ec_sys with write support though, right? I want to use secure boot, which is why I am moving to this module. Every time my battery runs out, the bios resets or forgets the secure boot disabled setting and I am tired of it.

Which signing keys you'll use? Own? Will they be reset after battery will die? I'm not sure is module need to be signed or you need only the signed loader like GRUB from Ubuntu to use stock keys

heeen:
I disabled secure boot via the bios/security tab.
Now  I enrolled the dkms signing key ubuntu created at `/var/lib/shim-signed/mok/MOK.der` so hopefully all dkms built modules are signed automagically. it appears to work for this module and nvidia which are the two modules I care about right now.

teackot:
Merged

Issue #138, opened 2024-07-02
GF76 11-UC


Waujito:
### Laptop model

MSI Katana GF76 11UC

### EC firmware version

17L2EMS1.108

### EC memory dump

```text
| _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 00 04 0b 0b
0x3_ | 03 00 00 0d 00 00 50 81 00 00 00 00 00 00 00 00
0x4_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 2f 00 37 40 49 4c 52 58
0x7_ | 64 26 26 2b 30 36 3c 46 55 64 08 03 03 03 03 03
0x8_ | 00 00 37 3d 43 49 4f 54 63 00 00 2b 30 36 3c 46
0x9_ | 55 64 08 03 03 03 03 02 02 0f 7d 02 0a 78 39 00
0xa_ | 31 37 4c 32 45 4d 53 31 2e 31 30 38 30 34 31 30
0xb_ | 32 30 32 33 31 33 3a 34 34 3a 34 32 00 00 00 28
0xc_ | 00 00 07 22 00 00 00 00 00 d1 00 00 00 00 00 00
0xd_ | 00 00 c1 83 0d 00 05 80 00 01 00 00 00 00 00 00
0xe_ | e2 00 00 00 00 00 00 40 00 00 00 00 00 c0 00 00
0xf_ | 40 00 70 00 00 64 00 00 64 00 00 00 00 00 00 00
```

### GPU

Nvidia

### Is your keyboard RGB?

No (single color)

### Additional context

I will try to provide support for it with myself but I will probably need in your help.
So far I have found:
✔️ Cooler boost
✔️ Webcam toggle 
✔️ Webcam block 
✔️ Fn <-> Win 
✔️ Mic mute LED 
✔️ Sound mute LED 
✔️ Keyboard backlight intensity

❓ Shift mode
❓ Fan mode

[...]

PR #139, opened 2024-07-02
Support for Katana GF76-11UC, 17L2EMS1.108

Waujito:
Support Katana GF76-11UC (#138)

List of changes: 
- **[Support for reverse fan speed.](https://github.com/BeardOverflow/msi-ec/pull/139/commits/0bd5dae9fda187ed9241e17a32ae42b913eb9502)**  
On tihs platform fan speed is reversed (more value in ec -> less fan speed). This PR adds a workaround for this problem without break of existing APIs.  
A one problem keeps: In fact fan speed takes 2 bytes (for me, 0xc8 and 0xc9) but the module supports only one byte (0xc9). 
On my laptop CPU fan speed is inversed so when fan gets slow it turns out to be (0x00 0xfa -> 0x00 0xff -> 0x01 0x08) Which may lead to EINVAL (when it reads 0x08 and compares it with maximum speed). But this is not critical because such a big value is reported for about 1 second and only when fan is about to stop and become 0x00 0x00.
- **[Fix keyboard brightness reset on module exit.](https://github.com/BeardOverflow/msi-ec/pull/139/commits/e579abfeaf42f74d8a2f7c6869453ac8bb1c1c31)**
By default kernel triggers the setter on led_classdev_unregister which
leads to keyboard brightness reset (the keyboard backlight turns off).
This behavior may cause resets every system reboot. Also it's quite
annoying in development.

### DIFF
--- a/ec_memory_configuration.h
+++ b/ec_memory_configuration.h
@@ -63,11 +63,18 @@ struct msi_ec_fan_mode_conf {
 	struct msi_ec_mode modes[5]; // fixed size for easier hard coding
 };
 
+#define FAN_PARSE_STRATEGY_NORMAL 0
+// When fan speed specified in reverse order 
+// 0xff - minimal speed, 0x01 - maximum and 0x00 - stopped.
+#define FAN_PARSE_STRATEGY_REVERSE 1
 struct msi_ec_cpu_conf {
 	int rt_temp_address;
 	int rt_fan_speed_address; // realtime
 	int rt_fan_speed_base_min;
 	int rt_fan_speed_base_max;
+	// For special fan speed parsing cases
+	// Optional parameter. Defaults to FAN_PARSE_STRATEGY_NORMAL.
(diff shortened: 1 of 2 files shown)


glpnk:
You shouldn't use RPM values, this thing is broken by design.

Technically, you can add RPM values and read them separately. This also will require separating %RPM and RPM addresses

Waujito:
As the driver for support fan speed in sensors is out, I think this PR is obsolete. The only thing I want to keep is [Fix keyboard brightness reset on module exit.](https://github.com/BeardOverflow/msi-ec/pull/139/commits/e579abfeaf42f74d8a2f7c6869453ac8bb1c1c31) commit because it is annoying to see light swithces for keyboard backlight when the driver exists.

teackot:
> The only thing I want to keep is `Fix keyboard brightness reset on module exit.` commit

Alright, I cherry-picked it. And added a docs entry for Katana GF76 11UC

PR #143, opened 2024-07-09
Implement Fan Thermal Curve

Waujito:
Hi!
I hope this PR will be helpful as a reference implementation of the fan thermal curve in further development.
So, I got tired of always having hot keys, unstable MControlCenter (it seems to have a tendency to switch to msi-ec, so it needs this feature too) and had implemented the curve myself in this module. This implementation still needs a lot of testing and is only available for my laptop now, but I think there is enough information to port it to other devices (idk about WMI1 but WMI2 seems to be ok). So, how it is:
The module creates `/cpu/curve` and `/gpu/curve` attributes, which are plain text files with curve data. The data is stored in the following order:
`s0 t1 s1 t2 s2 ... t(n-1) s(n-1) t(n) s(n)\n` where s* is a fan speed, t* is a temperature. Note that there is no t0 because it is `less than t1` (in fact there is likely a field for t0 in EC, which defaults to 0 but I think the behavior for temperatures less than it is undefined, so it is safer to just omit it). Previously I used to think that a more sysfs-ish way to represent the curve is a separate attribute group with `s0 t1 ... sn tn` attributes but this involves partial updates, while the user expects atomicity. Also there were thoughts about bin_attribute with curve representation in binary format (without spaces and each character represents curves parameter) but this involves too much overhead and is hard to read/write for the user.

Also about the implementation: It is different from other attributes because it is stored in memory and syncs with EC lazily. I did this because of #138 where we confirmed that the fan speed is broken on some devices when a non-default curve is in EC in non-advanced fan_mode. So on such devices curve fallback to default values is available: in non-advanced mode the curve keeps its default value, but the end user can virtually change it. /*/curve attributes are available but do not affect EC (they goes on when advanced mode is enabled).

### DIFF
--- a/ec_memory_configuration.h
+++ b/ec_memory_configuration.h
@@ -63,6 +63,32 @@ struct msi_ec_fan_mode_conf {
 	struct msi_ec_mode modes[5]; // fixed size for easier hard coding
 };
 
+/**
+ * Curve maximum entries (should be more than real maximum for extensibility).
+ */
+#define CURVE_MAX_ENTRIES 16
+
+/**
+ * Persists curve in 
+ */
+#define CURVE_APPLY_STRATEGY_NORMAL 0
+
+/**
+ * Resets curve from ec to default when auto mode is turned on.
+ * Required on some devices where auto mode is broken by custom curve in EC.
+ */
+#define CURVE_APPLY_STRATEGY_RESET_ON_AUTO 1
+
+// Curve start address and entries count.
+struct msi_ec_fan_curve {
+	int speed_start_address;
+	int temperature_start_address;
+	int entries_count;
+
+	// Defaults to CURVE_APPLY_STRATEGY_NORMAL
+	int apply_strategy;
+};
+
 struct msi_ec_cpu_conf {
 	int rt_temp_address;
 	int rt_fan_speed_address; // realtime
@@ -71,11 +97,14 @@ struct msi_ec_cpu_conf {
 	int bs_fan_speed_address; // basic
 	int bs_fan_speed_base_min;
 	int bs_fan_speed_base_max;
+
+	struct msi_ec_fan_curve fan_curve;
 };
 
 struct msi_ec_gpu_conf {
 	int rt_temp_address;
 	int rt_fan_speed_address; // realtime
+	struct msi_ec_fan_curve fan_curve;
 };
 
 struct msi_ec_led_conf {
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -3344,7 +3344,7 @@ static int curve_fan_mode_change(const char *mode) {
 		for (int i = 0; i < ALL_CURVES_COUNT; i++) {
 			struct curve_pack curve_data = *all_curves_list[i];
 
-			if (conf.cpu.fan_curve.apply_strategy == CURVE_APPLY_STRATEGY_RESET_ON_AUTO &&
+			if (curve_data.curve->apply_strategy == CURVE_APPLY_STRATEGY_RESET_ON_AUTO &&
 				is_curve_allowed(*curve_data.curve)) {
 
 				int status = push_ec_curve(*curve_data.curve, 
@@ -3360,7 +3360,7 @@ static int curve_fan_mode_change(const char *mode) {
 		for (int i = 0; i < ALL_CURVES_COUNT; i++) {
 			struct curve_pack curve_data = *all_curves_list[i];
 
-			if (conf.cpu.fan_curve.apply_strategy == CURVE_APPLY_STRATEGY_RESET_ON_AUTO &&
+			if (curve_data.curve->apply_strategy == CURVE_APPLY_STRATEGY_RESET_ON_AUTO &&
 				is_curve_allowed(*curve_data.curve)) {
 
 				sync_ec_curve_safe(*curve_data.curve, curve_data.curve_fan_speed, curve_data.curve_temp);
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -57,6 +57,7 @@
 #define FM_BASIC_NAME		"basic"
 #define FM_ADVANCED_NAME	"advanced"
 
+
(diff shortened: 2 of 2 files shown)


heeen:
this PR appears to work for my 16V5EMS1.108 GS66-12UHS. as for the interface, maybe a leading t0=0 in the curve file would make it easier to read for humans

Waujito:
I have been using the curve for about 2 months and can say that it works fine. The only noticeable thing is curve override after suspend. But I don't think this is module's problem, as we can easily fix it in userspace, like this:
```
[Unit]
Description=Restart Bluetooth after resume
After=suspend.target

[Service]
Type=simple
ExecStart=/bin/systemctl --no-block restart custom_curve.service

[Install]
WantedBy=suspend.target
```

custom_curve is just a service that pushes the curve to msi-ec.

About that leading t0=0 I think I can add it, but the driver should check it to be 0 always.

heeen:
would you mind posting your systemd service files?

Waujito:
> would you mind posting your systemd service files?

They are just simple scripts to set up and off the curve:
 /etc/systemd/system/custom_curve.service
```
[Unit]
Description=CustomCurve

[Service]
Type=oneshot
ExecStart=/bin/curve_apply
RemainAfterExit=true
ExecStop=/bin/curve_stop
ExecReload=/bin/curve_apply

[Install]
WantedBy=multi-user.target
```

/etc/systemd/system/cscurve_suspend.service
```
[Unit]
Description=Reset curve after resume
After=suspend.target

[Service]
Type=simple
ExecStart=/bin/systemctl --no-block restart custom_curve.service

[Install]
WantedBy=suspend.target
```

/bin/curve_apply 
```
#! /bin/bash
set -e
curve="0 50 41 55 55 60 65 65 85 68 100 80 120"
echo $curve > /sys/devices/platform/msi-ec/cpu/curve
echo $curve > /sys/devices/platform/msi-ec/gpu/curve
echo advanced > /sys/devices/platform/msi-ec/fan_mode
```

/bin/curve_stop
```
#!/bin/bash
set -e
echo auto > /sys/devices/platform/msi-ec/fan_mode
```

mirek967:
Waujito, do you plan to update the patch up to current version of msi-ec?

Waujito:
Yeah, the full year after the first commit, and I decided to update. I've been using this curve functionality all the year and it really works perfect. Today I have updated the code (basically rewritten), so this may have some hidden bugs. But after the review, I am pretty sure this should be included in the main branch of msi-ec. Please note, that this is implemented for and tested on only my firmware version: `17L2EMS1.108`. For other firmware, user should explicitly provide .fan_curve config. I have no such experience with ec on other MSI laptops to provide curve configs for them, so I ask maintainers to help with it. But we can just accept this PR and provide support for other firmware on demand later.

> maybe a leading t0=0 in the curve file would make it easier to read for humans

Yeah, may be. Let's wait for maintainer's thoughts about it.

As this PR works well, I mark it as ready for review and merge.

mirek967:
Thank you Вадим.

Waujito:
Ready for review @glpnk

mirek967:
I added it into CONF28, msi Titan 18HX. It have the same count of point as Vadim setting, and the same addresses. 
Works well.

glpnk:
@Waujito Hi. I will test your new functionality.

Currently Bazzite devs upstreaming fan curve API for `msi-wmi-platform` driver. Can you test it? It's compliant with `hwmon` API but some features are unclear for me. They've also implemented fan curve caching.

glpnk:
Your Katana GF76 11UC is WMI2 laptop for which this module is intended https://github.com/bazzite-org/patchwork/commits/msi/

Waujito:
> @Waujito Hi. I will test your new functionality.
> 
> Currently Bazzite devs upstreaming fan curve API for `msi-wmi-platform` driver. Can you test it? It's compliant with `hwmon` API but some features are unclear for me. They've also implemented fan curve caching.

Wow, this is interesting. They've found an ACPI interface allowing to control the curve without ec.  And it works! 

I found one problem: it does not allow the fans to reach the maximum speed: msi-wmi-platform caps it in 100% while 150% seems to be the real maximum, but my fans reach their maximum speed at 120% :)

To make the curve caching work I've added a DMI_MATCH in msi_quirks.

glpnk:
Well, history of this driver started in `lm-sensors` issue thanks to @wer-wolf https://github.com/lm-sensors/lm-sensors/issues/475

> I found one problem: it does not allow the fans to reach the maximum speed: msi-wmi-platform caps it in 100% while 150% seems to be the real maximum, but my fans reach their maximum speed at 120% :)

Some old laptops had 0-100% range (?), newer laptops have 0-150% with unclear relation of % to RPM. Not sure what Linux Fan APIs need 0-255 range or any other is okay.

> Wow, this is interesting. They've found an ACPI interface allowing to control the curve without ec. And it works!

It's writing to EC through Linux ACPI-WMI stack.

---

I know that some features have support bits, which we can read to disable certain modes and features. But some of them are broken. For example - MSI have laptop series which have both RGB and monochrome KB BL, and backlight support bit is set on them.

But mute leds not have support bits. Number of fans is hardcoded in text file distributed with windows apps.

---

Also some user requested support for device, which have non standard EC layout. But WMI2 emulates certain parts of original EC. #186 

UPD: because app uses both namespaces and raw addresses, msi devs mapped both.

---

Also some interesting document about structure was found https://github.com/Sparronator9999/YAMDCC/issues/89

Waujito:
> I know that some features have support bits, which we can read to disable certain modes and features. But some of them are broken. For example - MSI have laptop series which have both RGB and monochrome KB BL, and backlight support bit is set on them.

Oh yeah chineese technologies 😁

mirek967:
> > I know that some features have support bits, which we can read to disable certain modes and features. But some of them are broken. For example - MSI have laptop series which have both RGB and monochrome KB BL, and backlight support bit is set on them.
> 
> Oh yeah chineese technologies 😁

Like all msi computers we own.

PR #152, opened 2024-08-19
Add guide for address discovery

mutchiko:


glpnk:
@teackot what can you say?

teackot:
ec_sys isn't present on all distributions, for example Fedora doesn't ship it. We have debug mode in our driver, which dumps the EC ram contents in `/sys/devices/platform/msi-ec/debug/ec_dump`

glpnk:
Hi, @teackot, can you check this EC dump method on your devices with both generations. It should workon WMI1 devices, because used in DSDT tables. It reads address `0xFC000080`
```
sudo dd if=/dev/mem bs=1 skip=4227860480 count=256 | hexdump -C
```

teackot:
Hi! Yep it works on my WMI2 Summit E14, gives me a correct EC dump. I'll test in on my P14 a little later

glpnk:
Nice, I'll add this to guide too

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+until you change the settings in the MSI app.
+
+Example: To figure out which **Address** is used by user scenario (`shift
+mode`) go to the user scenario page in the MSI app and keep changing it
+while looking at the EC table.
+
+Eventually you'll notice a **Value** that changes each time you change the
+setting (or maybe two **Values** in two **Addresses** that change at the same
+time), once you find the **Address**, you can start writing down the **Value**
+for EACH user scenario, so you can report it.
```
teackot, line 71:
```suggestion
Example: To figure out which **Addresses** are used by user scenarios (or shift
modes) go to the user scenario page in the MSI app and keep switching the scenarios
while looking at the EC table.

Eventually you'll notice one or more values that change each time you change the
setting, once you find them, you can start writing down the addresses and values
corresponding to *each* user scenario, so you can report them.
```

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
+# Intro
+To get your MSI laptop supported, there are two main methods, one that
+requires windows to be installed, and the other works directly on linux.
+
+If you have any bios/firmware updates on the official MSI website, and
+you haven't updated yet then **DO NOT UPDATE!** follow the guide first:
+
+## Windows method(recommended):
```
teackot, line 8:
```suggestion
## Windows method (recommended):
```

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+time), once you find the **Address**, you can start writing down the **Value**
+for EACH user scenario, so you can report it.
+
+### Linux method
+
+***
+
+This method has limited results compared to the Windows method, because most
+MSI laptop features are tied to software toggles that can be only found in their apps
+installed on windows.
```
teackot, line 79:
```suggestion
This method is very limited compared to the Windows method, because most
MSI EC features are tied to software toggles that can only be found in the official Windows apps.
```

teackot (commented):
I left some suggestions on how to improve the guide

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
+# Intro
+To get your MSI laptop supported, there are two main methods, one that
+requires windows to be installed, and the other works directly on linux.
```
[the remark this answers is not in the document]
mutchiko: it does.

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+
+If you have any bios/firmware updates on the official MSI website, and
+you haven't updated yet then **DO NOT UPDATE!** follow the guide first:
```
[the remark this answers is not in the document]
mutchiko: i went for the short route because the guide was meant to get people started with as little side info as possible.

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+
+* `cat /sys/kernel/debug/ec/ec0/io > ec.dump`
+
+then you can send us the output, and wish us luck.
+
+
+P.S: you should repeat the method you used after EACH BIOS/Firmware update, the reason you
+shouldn’t update before following the guide is that the values and addresses could change after
+updating, and it is important to support all versions of the BIOS/Firmware or the driver might
+not work after the update.
```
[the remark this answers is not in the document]
mutchiko: curiosity was supposed to make the reader go through the guide from start to finish

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+7.  Launch it as administrator:
+![run as admin](pics/support_guide/run_as_admin.png)
+
+8. Navigate to the EC tab (page):
+![open ec tab](pics/support_guide/open_ec_tab.png)
+
+9. Here you should see a table of all the values your Embedded Chip has in its memory.\
+The values you see can be changed manually (by writing to them)\
+\
+DO NOT DO THAT: writing the wrong the values to the wrong address might brick the laptop completely and EC/BIOS RESET CAN'T FIX THAT!<br/>
```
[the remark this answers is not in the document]
mutchiko: maybe being straightforward is not the best way to keep people from messing up, i prefer it my way because it sounds scarier. but im not against the change.

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+![not apply changes](pics/support_guide/not_apply_changes.png)
+
+10. Change the reading speed to 500-600ms,
+this makes it easier to see how the
+values react to the MSI app settings:
```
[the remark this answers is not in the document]
mutchiko: i agree

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+![refresh rate button](pics/support_guide/refresh_menu.png)
+
+11. Reading addresses: lets say you are looking for a specific address
+0x54 (0xFirstNumberSecondNumber) FirstNumber can be found
+on the left side of the table and SecondNumber can be found on
+the top side of the table:
```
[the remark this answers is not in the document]
mutchiko: this is way better, but lets call them numbers instead of digits, just to match it with the visual representation in the screenshot.

docs/device_support_guide.md
```diff
@@ -0,0 +1,108 @@
...
+![hex editor how to](pics/support_guide/hex_editor_how_to.png)
+
+Each **Address** contains a **Value**, when you locate an **Address** inside the
+table you can write it like this 0x54 = 00 (**Address** = **Value**)
+This **Value** can change depending on what it is related to, if you find the
+**Address** for CPU temperature; the **Value** will change on its own, if you
+find the **Address** for battery charge limit, then the **Value** won't change
+until you change the settings in the MSI app.
```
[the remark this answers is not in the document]
mutchiko: shorter paragraphs are indeed easier to read

docs/device_support_guide.md
mutchiko: i abused words in bold here because i wanted it to be easier to understand

docs/device_support_guide.md
mutchiko: how do people know if a feature is actually an EC feature and not related to driver or software solutions on windows? lets keep the more generalized sentence on this one and we might avoid some confusion.

docs/device_support_guide.md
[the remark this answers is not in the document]
teackot: Mention of the ability to write the data into the EC should definitely be fused together with the warning, otherwise the first paragraph by itself reads "here's the data and you can edit it".

I think making the *whole* warning bold and the damage&brick part ALL CAPS would be scary enough (I edited the suggestion)

docs/device_support_guide.md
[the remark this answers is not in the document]
teackot: Alright, sounds good

docs/device_support_guide.md
teackot: Sounds good!

docs/device_support_guide.md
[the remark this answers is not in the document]
glpnk: Github have this and similar tags, so ve can use them to highlight some dangerous things

> [!IMPORTANT]
qq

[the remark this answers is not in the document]
teackot: Caution looks like the right one, let's see

> [!CAUTION]
**Do not attempt to edit the memory manually**. Writing the wrong the values to the wrong addresses will **DAMAGE YOUR HARDWARE** or even **BRICK YOUR DEVICE**! This can't be reversed by resetting the EC or BIOS.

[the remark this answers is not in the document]
mutchiko: i just refreshed the page and the first thing that caught my eyes are the words in bold, lets keep the caution section because it looks good, but its not as eye catching as BIG **BOLD** words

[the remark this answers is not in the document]
glpnk: Nice.

glpnk:
@teackot I'm thinking about splitting guide into 2 or 3 parts:
+ main - Disclaimers, info about known values, other info
+ Windows - Disclaimer + Windows guide
+ Linux - Disclaimer + debug mode + ec_sys + raw memory read

teackot:
So the main one will describe the base principle, the danger, info about the dump, and where and what to look for (aka the known values), etc? This does sound like it should be in this guide.

If by splitting you mean to split it into separate files I don't think there is much need and a single page might be better

I'm thinking of a format like this:

1. Base info (main guide)
2. Table of contents
3. Windows
4. Linux

glpnk:
Okay, known values is for contribution guide, for example.

> I'm thinking of a format like this:
>
>    Base info (main guide)
>    Table of contents
>    Windows
>    Linux

Okay, single file guide sounds good. But what we can add to `Table of contents`?

Also, can you start new review and add [Alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) tags

My own guide making attempt here https://github.com/glpnk/msi-ec/tree/docs/more-docs/docs

teackot:
We can put the known values into its own section, since it is likely going to be big.

- Base info
- Table of contents
- Windows
- Linux
- Known values

All sections after the TOC go into it (win, lin, known)

docs/device_support_guide.md
```diff
@@ -0,0 +1,99 @@
...
+
+7.  Launch it as administrator:
+![run as admin](pics/support_guide/run_as_admin.png)
+
+8. Navigate to the EC tab (page):
+![open ec tab](pics/support_guide/open_ec_tab.png)
+
+9. Here you should see a table of all the values your Embedded Chip has in its memory.
+
+**Do not attempt to edit the memory manually. Writing the wrong the values to the wrong addresses will DAMAGE YOUR HARDWARE or even BRICK YOUR DEVICE! This can't be reversed by resetting the EC or BIOS.**
```
teackot, line 40:
```suggestion
> [!CAUTION]
**Do not attempt to edit the memory manually. Writing the wrong the values to the wrong addresses will DAMAGE YOUR HARDWARE or even BRICK YOUR DEVICE! This can't be reversed by resetting the EC or BIOS.**
```

docs/device_support_guide.md
```diff
@@ -0,0 +1,99 @@
+# Intro
+There are two main methods to get your MSI laptop supported: the recommended method requires Windows to be installed, and the other works directly on Linux.
+
+If there are any BIOS/firmware updates available for your laptop, follow this guide for your current firmware before installing any of them. Then repeat the process for each new firmware version you install from the official MSI website. This is required to obtain support for older firmware as the EC configuration may vary across the versions.
```
teackot, line 4:
```suggestion
> [!IMPORTANT]
If there are any BIOS/firmware updates available for your laptop, follow this guide for your current firmware before installing any of them. Then repeat the process for each new firmware version you install from the official MSI website. This is required to obtain support for older firmware as the EC configuration may vary across the versions.
```

teackot (commented):
Alerts

glpnk:
Main problem of Markdown is indents in lists
image

teackot:
We can remove the "9. Here you should see a table of all the values your Embedded Chip has in its memory." entry and put the alert it the "reading the table" section.

It can go like this:

```
## Windows method

### Preparation

1. Install
2. Update
3. MSI apps
4. AMD/Nvidia
5. Test the app
6. Download RWEverything
7. Launch
8. EC page
9. Refrest rate

### Reading the table

Here you will see a table of all the values your Embedded Controller has in its memory.

CAUTION: do not edit

This is how you find the address

This is how you interprete it

Example

...
```

This way the list won't have anything that breaks the indentation

glpnk:
Yeah, because we have second method of reading EC on Win we can split guide to few steps as you said, and apps

docs/device_support_guide.md
```diff
@@ -81,21 +79,83 @@ If you are lucky, your laptop model will have similar addresses to another lapto
...
+
+### Reading the EC RAM mapped to system memory
+
+Some devices have EC memory mapped to system memory address `0x0xFC000800`,
+so you can read it from the `/dev/mem` pseudo file with `dd`.
+
+`/dev/mem` may not be supported on some distros. For details, see `man mem`.
+
+> [!WARNING]
+> Reading random parts of the system memory can reveal your secrets, so check the dump before you post it to Github.
```
mutchiko, line 172: and how do i check if the dump i made doesn't include sensitive info? let's call this method "experimental" until we make sure it works no less than 90% of the time.

glpnk: true, only by comparing with reference dump, like matching the location of the ec name

glpnk: Actually, chance of getting reading error is higher than accidentally dump some confidential info. And maybe this memory area newer mapped to system RAM

docs/device_support_guide.md
```diff
@@ -4,7 +4,30 @@ There are two main methods to get your MSI laptop supported: the recommended met
 > [!IMPORTANT]
 > If there are any BIOS/firmware updates available for your laptop, follow this guide for your current firmware before installing any of them. Then repeat the process for each new firmware version you install from the official MSI website. This is required to obtain support for older firmware as the EC configuration may vary across the versions.
 
-## Windows method (recommended):
+# Table of contents (rough)
+
++ FW naming and WMI versioning explanation(?) (TODO)
```
mutchiko, line 9: this is outside the scope of this guide, you've already done a good [write up](https://github.com/BeardOverflow/msi-ec/discussions/98) on that topic, so lets keep it outside this guide

teackot: Right, this is just a guide on how to obtain and interpret the dump

docs/device_support_guide.md
teackot: Let's mark it as "Not recommended" and put the warning at the top of this method's section, so it can be the last resort if the two previous methods didn't work for some magical reason

docs/device_support_guide.md
```diff
@@ -0,0 +1,184 @@
...
+
+* `hexdump -C /sys/kernel/debug/ec/ec0/io`
+
+Or save it to file:
+
++ `hexdump -C /sys/kernel/debug/ec/ec0/io > ec_dump.txt`
+
+If you need dump in binary form:
+
+* `cat /sys/kernel/debug/ec/ec0/io > ec_dump.bin`
```
teackot, line 162: I don't think anyone does. At least, we don't need binaries submitted here, so let's exclude this part

docs/device_support_guide.md
teackot: Alright, adding it here as a draft wouldn't hurt, we'll see how well it fits here when the guide is near finished

glpnk:
Sort of ready to review

docs/device_support_guide.md
```diff
@@ -163,3 +165,39 @@ To read EC memory in text form run:
...
+    + I - Intel CPU
+  + For EC:
+    + E - ENE
+    + I - ITE
++ ${\color{yellow}n}$ - EC / board(?) model
++ ${\color{WildStrawberry}y}{\color{YellowOrange}zz}$ - version
+  + ${\color{WildStrawberry}y}$ - board generation; firmware is incompatible if different
+  + ${\color{YellowOrange}zz}$ - firmware version
+
+## Firmware generations
```
mutchiko, line 191: i don't see how this might help people discover addresses on their unsupported laptops

docs/device_support_guide.md
```diff
@@ -163,3 +165,39 @@ To read EC memory in text form run:
 To save EC memory in text form run:
 
 + `sudo dd if=/dev/mem bs=1 skip=4227860480 count=256 | hexdump -C > ec_dump.txt`
+
+# Contribution
+
+## Firmware naming
```
mutchiko, line 171: this is a solid explanation , but it's better to put it in another documentation

glpnk: We probably don't need more then 1 guide now

docs/device_support_guide.md
glpnk: If people will compare with reference map they need to know how to check about generations

glpnk:
@teackot do you have access to https://github.com/BeardOverflow/msi-ec/graphs/traffic ? Can you make screenshot of `Referring sites` block? Maybe this will help us focus on the most popular sites that bring people to this project.

teackot:
Here it is

image

Most visits are from search engines and Phoronix articles

docs/device_support_guide.md
teackot: These should probably be in a separate guide.

Some people have been requesting a guide on adding a config to the code and I think it also doesn't belong in this file.

We can make a file with a "Contribution" guide that:

1. refers people to this guide if they want to contribute by discovering their laptops' configs (this is a separate file because it is big)
2. provides some specific info like FW naming and generations
3. explains how to add a config to the code

teackot:
What's left to do here? I think we need to direct people to the issue tracker at the end of the "Windows method" section (like we do in the "Linux method" section) and remove the "Contribution" section (to later move it to a different file as described in [this](https://github.com/BeardOverflow/msi-ec/pull/152#discussion_r1730415430) comment), and this PR is good to go

mutchiko:
@teackot only thing left is is to decide where to put links to the guide, my initial thoughts:
1- in the first/second section of the readme
2- on top of the supported devices list
3- in the issue template

and this is just enough to attract people's attention.

glpnk:
Now I think that `Table of content` section looks confusing, we probably need to move 2nd level bullet points to guide.

About contribution section - I think info about generations might be useful, but we can call it like `Additional info`

glpnk:
> 2- on top of the supported devices list

image

mutchiko:
@glpnk for number 2 lets change the wording to "If your device is not in the list, read this _guide_ to get it supported and add it to the list"

helpful in case people outside github share the link with each other

glpnk:
I'm not committed this change, so it's just draft

mutchiko:
Im thinking about adding an explanation about mic/speaker mute LEDs in the Additional Info section;
just to tell people that they exist and how they work on linux, but we don't have a way to trigger them, yet.

glpnk:
LEDs controlled by kernel triggers, which are controlled by audio card drivers. Bluetooth audio works in userspace, so it can't be used as trigger

mutchiko:
@teackot the guide is ready, please go ahead and merge it.

teackot:
Thank you very much for writing this important guide!

teackot:
And with this, all the open PRs are finally merged

PR #161, opened 2024-09-19
fix type to compile for Kernel 6.11

arunk140:
6.10 - https://www.kernel.org/doc/html/v6.10-rc3/driver-api/driver-model/platform.html#platform-drivers

```
 int (*remove)(struct platform_device *);
```
changed to

```
void (*remove)(struct platform_device *);
```
in the latest Linux kernel (6.11)

6.11 - https://www.kernel.org/doc/html/latest/driver-api/driver-model/platform.html#platform-drivers

```
struct platform_driver {
      int (*probe)(struct platform_device *);
      void (*remove)(struct platform_device *);
      void (*shutdown)(struct platform_device *);
      int (*suspend)(struct platform_device *, pm_message_t state);
      int (*resume)(struct platform_device *);
      struct device_driver driver;
      const struct platform_device_id *id_table;
      bool prevent_deferred_probe;
      bool driver_managed_dma;
};
```

### DIFF
--- a/6.11-kernel.patch
+++ b/6.11-kernel.patch
@@ -1,0 +1,22 @@
+diff --git a/msi-ec.c b/msi-ec.c
+index 830cdf4..80ce756 100644
+--- a/msi-ec.c
++++ b/msi-ec.c
+@@ -3562,7 +3562,7 @@ static int msi_platform_probe(struct platform_device *pdev)
+ 	return sysfs_create_groups(&pdev->dev.kobj, msi_platform_groups);
+ }
+ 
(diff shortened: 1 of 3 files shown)


glpnk:
As 6.11 kernel is not yet mainstream in some distros, like Ubuntu, consider make patch file and add guide to README.

Thanks for your findings.

arunk140:
Yep makes sense. Updated the commit with the patch and README changes.

glpnk:
Thanks for your commit

teackot:
Thanks!

PR #168, opened 2024-10-05
msi-ec: add support for kernel 6.11

quaresmajose:


teackot:
I'll also merge the other patch into the code, it would be more reliable than having to maintain multiple patches for older APIs

glpnk:
Like 6.2 patch? I remember you previously found a reason to make <6.2 patch rather than macro

teackot:
Mainly to keep the code clean, but it makes it harder to contribute for people running older kernels as they'd have to apply the patch to build the code and then revert it back to commit. The patches can also potentially break if the code surrounding them changes

PR #131, opened 2024-06-07
Add MSI GF63 Thin 9SC support

smalltheif:
EC Firmware version: `16R3EMS1.104`

### Issues :
- the reported positions of the FN and Win keys are inverted
- after each reboot, the shift mode is back to `unspecified`
- the EC dumps are unchanged for any keyboard backlighting modes

### EC dump:
```
     | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 80 00 00 00 00 00 00 00 00 00 14 00 00 14 00
0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x2_ | 00 14 00 00 00 00 00 00 00 00 00 80 c0 04 78 0b
0x3_ | 03 89 00 0d 50 0a 37 00 94 11 88 2c c2 01 e0 48
0x4_ | 00 00 64 00 e5 0f 00 00 e5 0f 69 32 ea 0b 00 00
0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 37 00 37 40 49 4c 52 58
0x7_ | 64 23 23 26 2a 2d 32 37 3e 00 08 03 03 03 03 03
0x8_ | 2e 00 37 3d 43 49 4f 55 63 23 23 26 2a 2d 32 37
0x9_ | 3e 00 08 03 03 03 03 03 02 00 64 02 00 6e 00 00
0xa_ | 31 36 52 33 45 4d 53 31 2e 31 30 34 ad 39 31 34
0xb_ | 32 b2 32 30 31 30 3a 31 37 3a 33 32 e5 0f 40 c0
0xc_ | 00 07 25 00 00 00 46 00 00 75 00 00 00 e2 14 00
0xd_ | 00 00 00 80 7f 00 d1 00 00 00 00 00 00 00 00 00
0xe_ | e2 02 00 fa 10 00 00 00 40 00 00 00 00 d1 00 e4
0xf_ | 00 00 80 83 0d 23 05 f5 00 00 00 00 00 00 00 00
```

### Issues resolved:
- [#104](https://github.com/BeardOverflow/msi-ec/issues/104)
- [#118](https://github.com/BeardOverflow/msi-ec/issues/118)

PR #172, opened 2024-10-28
Code cleanup before implementing Fan curve HWMON API

glpnk:
Checklist:

- [x] Remove Basic Fan file because it's messed. Basic fan related issues https://github.com/BeardOverflow/msi-ec/pull/131#issuecomment-2155556926 https://github.com/YoyPa/isw/issues/204#issuecomment-904012572
- [ ] Add/Remove GPU fan addresses according to file from Dragon Center app

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,12 +174,7 @@ This driver exports a few files in its own platform device, msi-ec, and is avail
 - `/sys/devices/platform/msi-ec/cpu/realtime_fan_speed`
   - Description: This entry reports the current cpu fan speed.
   - Access: Read
-  - Valid values: 0 - 100 (percent)
-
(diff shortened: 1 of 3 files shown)


glpnk:
@teackot Hi, what you think about fan curve number? In EC we have 7 points but MSI app show only 6. On some older devices when app toggled fan mode to "advanced" it set 6 curve %RPM values to zero (probably bug), but 7 was default, probably saving CPU from burning

teackot:
Hi

```c
	else if (attr == &dev_attr_cpu_basic_fan_speed.attr)
		address = conf.cpu.bs_fan_speed_address;
```
you need to remove the basic fan speed condition from is_visible()

glpnk:
Thanks. Fixed

PR #177, opened 2024-12-10
Small improvments for README and list of supported devices

mutchiko:
README:

- Mention that the driver can't fix ACPI errors
- Recommend using MCC as a graphical interface for msi-ec; new version will release [soon](https://github.com/dmitry-s93/MControlCenter/issues/203#issuecomment-2526208822) with this functionality supported by default.
- Mention Quick fixes discussion by @glpnk in the contribute section

List of supported devices:
- Recommend in-browser searching for laptop models to save time
- Recommend following the support guide in case the model is not supported yet

pull request is draft until everyone is OK with the changes and new MCC version is released


mutchiko:
New [MCC](https://github.com/dmitry-s93/MControlCenter) release is almost ready (0.5)

teackot:
Everything looks fine

MCC 0.5.0 was just released an hour ago, so I'm merging this

mutchiko:
@teackot thanks for the merge, i want to ask: do you use MCC yourself?

teackot:
No, using CLI was good enough for me. Also I'm maintaining this repo, so I need to use msi-ec to be aware if it breaks on some kernel update

(now that MCC supports msi-ec it might be useful to check it out)

mutchiko:
>was good enough for me

_was_? so you stopped using msi-ec?

teackot:
Oh, no, I just meant that I have only been using CLI until now, but might also add MCC to my workflow in the future, because it uses msi-ec now

mutchiko:
good to hear, im also planning on linking ```shift_mode``` with [power-profiles-daemon](https://gitlab.freedesktop.org/upower/power-profiles-daemon) in a way that changing the performance on one side will affect the other, i don't want to have a windows-like situation where there are many disconnected ways to change performance

teackot:
That's a good idea, I also thought about p-p-d but couldn't think of a right way to link it with the driver

mutchiko:
PPD has a Dbus solution, also i thought about executing regular CLI commands when changing modes in MCC.

i want to look at how desktop environments (KDE, gnome) implement it but i don't have the time

teackot:
> gnome

Definitely uses the dbus interface

On a [partially] unrelated note, have you noticed a problem with the micmute LED, which sometimes doesn't turn on when muting the microphone, even though `/sys/class/leds/platform::micmute/brightness` says 1? I wonder if it's a hardware problem or if it's related to the driver. This makes it quiet unreliable

mutchiko:
i don't think its related to the driver, i honestly gave up on making these tiny leds work on linux, but this https://github.com/BeardOverflow/msi-ec/discussions/130#discussioncomment-11010018 gave me some hope

muting the microphone never turns on the led for me (same with speakers), the only way is to write 1 manually to ```/sys/class/leds/platform::micmute/brightness``` but that doesn't depend on the state of the microphone.

i think what we are looking for is the correct triggers for these LEDs in pulsaudio/pipewire.

@glpnk knows better than me.

teackot:
I just checked, `...platform::micmute/brightness` says `1`, but the EC register is still `0x08` (the `off` state). I'll look into it later

> muting the microphone never turns on the led for me (same with speakers)

In my case micmute binds to pipewire, but speaker mute reqires @glpnk's fix

mutchiko:
i think there is some distro specific magic, what distro are you on? im on arch

teackot:
I'm on Fedora, but used to use Arch. IIRC neither of LEDs worked out of the box on Arch, so it might be related to the distro's default configs

mutchiko:
who knows, maybe speakers mute LED is one config line away from getting it to work

glpnk:
Stop calling audio led fix with my nickname, post has credits inside, haha

Issue #163, opened 2024-09-19
Please add new firmware for GL75 Leopard


mirek967:
### Laptop model

GL75 Leopard

### EC firmware version

17E8EMS1.101

### EC memory dump

```text
| _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
-----+------------------------------------------------
0x0_ | 00 00 00 46 00 00 2d 00 09 30 28 00 30 2d 0f c1
0x1_ | 00 12 13 4c 3e b0 96 46 19 30 1b 00 00 42 1f 4c
0x2_ | 21 37 00 00 00 00 00 00 00 00 2d 80 29 40 49 89
0x3_ | 03 09 01 05 58 2d 2f 38 54 0f 30 2a 88 01 e7 00
0x4_ | 00 00 61 00 08 0e 00 00 7d 0d 44 2f c8 0b 0c 30
0x5_ | 51 00 32 00 37 00 3e 58 00 00 00 00 00 00 00 00
0x6_ | 00 00 00 00 00 00 00 00 34 0a 32 38 3e 46 4e 56
0x7_ | 64 30 00 28 30 38 42 4c 56 00 0a 07 07 09 09 09
0x8_ | 2f 36 37 3c 41 46 4b 50 63 2d 00 2d 32 37 3e 46
0x9_ | 4e 00 0a 07 07 07 07 07 02 16 7d 02 16 61 00 00
0xa_ | 31 37 45 38 45 4d 53 31 2e 31 30 31 30 36 32 38
0xb_ | 32 30 32 30 31 36 3a 31 31 3a 31 38 0d 0e 00 40
0xc_ | 00 01 25 00 2d 9e b2 00 00 2d 00 b2 00 00 14 00
0xd_ | 00 00 20 00 7f 80 d1 00 86 00 00 40 00 18 09 13
0xe_ | e2 02 00 8a 10 00 00 00 00 00 00 00 00 83 00 d0
0xf_ | 00 00 c2 82 00 00 08 f8 00 00 00 c1 00 00 00 00
```

### GPU

Nvidia

### Is your keyboard RGB?

No (single color)

### Additional context

Proposed patch. I am testing it. It seems that everything works. 
On windows with dragon center I found also fan curves, but it seems right now they are not supported yet.
[leopard.zip](https://github.com/user-attachments/files/17065010/leopard.zip)

mirek967:
Is there any chance to get it included in near future?

[...]

PR #193, opened 2025-01-13
Add support for GL75 Leopard

mutchiko:
fixes issue https://github.com/BeardOverflow/msi-ec/issues/163


mutchiko:
need to update supported device list

teackot:
Merged

> need to update supported device list

74a839d

Issue #189, opened 2025-01-12
MSI Vector GP68HX 12VH Support


CemCanplt:
### Laptop model

MSI Vector GP68HX 12VH 

### EC firmware version

15M1IMS1.113-11272023-16:18:59

### EC memory dump

```text
Type:Embedded Controller: 0066,0062
Width:01
00=00 01=80 02=00 03=00 04=00 05=00 06=00 07=00 
08=00 09=00 0A=00 0B=00 0C=00 0D=00 0E=00 0F=00 
10=00 11=00 12=00 13=00 14=00 15=00 16=00 17=00 
18=00 19=00 1A=00 1B=00 1C=00 1D=00 1E=00 1F=00 
20=00 21=00 22=00 23=00 24=00 25=00 26=00 27=00 
28=0A 29=05 2A=00 2B=00 2C=00 2D=00 2E=0B 2F=5B 
30=03 31=01 32=00 33=0D 34=01 35=00 36=50 37=81 
38=2B 39=16 3A=28 3B=3C 3C=37 3D=02 3E=C0 3F=00 
40=2B 41=16 42=3C 43=00 44=2A 45=15 46=00 47=00 
48=A7 49=0C 4A=D3 4B=3D 4C=F9 4D=0B 4E=C0 4F=44 
50=00 51=00 52=00 53=00 54=00 55=00 56=00 57=00 
58=00 59=00 5A=00 5B=00 5C=00 5D=00 5E=00 5F=00 
60=00 61=00 62=00 63=00 64=00 65=00 66=00 67=00 
68=2F 69=00 6A=32 6B=39 6C=40 6D=46 6E=4C 6F=52 
70=64 71=28 72=00 73=28 74=30 75=3C 76=4B 77=59 
78=67 79=00 7A=0A 7B=08 7C=08 7D=08 7E=08 7F=08 
80=30 81=00 82=32 83=37 84=3C 85=41 86=46 87=4B 
88=64 89=30 8A=00 8B=30 8C=3C 8D=46 8E=52 8F=5D 
90=68 91=5F 92=07 93=07 94=07 95=07 96=07 97=07 
98=02 99=00 9A=00 9B=02 9C=00 9D=00 9E=00 9F=00 
A0=31 A1=35 A2=4D A3=31 A4=49 A5=4D A6=53 A7=31 
A8=2E A9=31 AA=31 AB=33 AC=31 AD=31 AE=32 AF=37 
[...]

PR #194, opened 2025-01-13
Adding MSI Vector GP68 HX 12V Config

CemCanplt:


teackot:
Merged

Bumped the conf id to 37 and also added a docs entry

teackot:
Closes #189

PR #195, opened 2025-01-14
Add Katana 15 B13VFK to supported devices

QkiZMR:


glpnk:
By the way, webcam hardware block can be read from software, not sure about ability to overwrite it. On devices without hardware toggle, you can lock button by writing to it

QkiZMR:
> By the way, webcam hardware block can be read from software, not sure about ability to overwrite it. On devices without hardware toggle, you can lock button by writing to it

It's not so important for me now. I'm planning to add after some time a new pull request with the address of the keyboard lights. So maybe I will look at the webcam block.

glpnk:
Rgb keyboards not supported, they are different device, in many cases, not related to EC, but part of USB keyboard, so try openrgb

glpnk:
Be careful, writing wrong values to keyboard breaks it

QkiZMR:
> Rgb keyboards not supported, they are different device, in many cases, not related to EC, but part of USB keyboard, so try openrgb

yes, keyboard is presents itself as USB device.

QkiZMR:
@glpnk When do you plan to merge changes?

glpnk:
@teackot

teackot:
I'll merge this and other remaining PRs today

teackot:
Merged!

Notes:
- bumped the config id to 36
- marked keyboard backlight as unavailable, as the EC interface is not supported by RGB keyboards
- removed the workaround note from the table because it is universal for all devices and we have it in the README

PR #221, opened 2025-02-20
added mutexes for functions that can race themselves

brachiGH:
added mutexes for  functions where a race condition could happen.


teackot:
Great, thank you. Now I can apply it to my patch set for the kernel

PR #228, opened 2025-02-22
Remove the battery_mode attribute

teackot:
It does the same thing that charge_control_(start/end)_threshold already do, just more restricted. Presets, if needed, should be provided by userspace utilities.


mutchiko:
MCC has a function to interact with the sysfs file, however, it is unused and instead uses the regular threshold limits.

ec seems to actually react and activates these profiles if you modify threshold to one of the profiles limits: 60-> min / 80->mid / 100->max

PR #266, opened 2025-04-03
msi-ec.c: Add Support for GF63-10SC

MasonT8198:
This seems to fully work on my GF63 running pop

auto fan mode actually autos which is nice, and turbo fan mode works too

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -883,6 +883,7 @@ static struct msi_ec_conf CONF11 __initdata = {
 };
 
 static const char *ALLOWED_FW_12[] __initconst = {
+	"16R5EMS1.102", // GF63 Thin 10SC-222
 	"16R6EMS1.104", // GF63 Thin 11UC
 	"16R6EMS1.106",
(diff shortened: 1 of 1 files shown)


MasonT8198:
way better than my cpu sitting at 90 consistently

MasonT8198:
Working
--------------
- Fan Modes
- Cooler Boost
- Webcam
- Shift Mode acts like it's working

Not Working
---------------
- KBD Backlight in Software (button still works ofc)
- Battery Charge Control
- no super battery at all
- fn_key or win_key functions

MasonT8198:
> @MasonT8198 No, this doesn't work this way, you mixing 2 different generations:
> 
> 
> 
> + `GF63 Thin 11UC` is `WMI2` generation
> 
> + `GF63 Thin 10SC` is `WMI1`
> 
> 
> 
> You can take any other `WMI1` device as reference. It will have majority of required features. You can break your EC by using wrong config

I'll just wait for the devs to add it in

PR #272, opened 2025-04-22
Add one shift mode for MSI Vector 16HX A14VHG

MaxenceTech:


MaxenceTech:
Eco/ Silent was added recently in MSI center

Issue #306, opened 2025-05-10
Support request for Prestige 14Evo B13M 14F1EMS1


gabrielmalaspina:
### Laptop model

Prestige-14Evo-B13M-243UK

### EC firmware version

14F1EMS1.207

### List of addresses and their values

```text
|      | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
|------+------------------------------------------------
| 0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 08 2c 09 08  |.............,..|
| 0x3_ | 02 05 00 05 00 00 50 81 c2 11 28 3c c8 01 c0 00  |......P...(<....|
| 0x4_ | bc 11 20 00 6b 0e ba fe 98 04 33 3a fe 0b c0 44  |....k.....3:...D|
| 0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x6_ | 00 00 00 00 00 00 00 00 2b 00 41 46 4b 50 55 5a  |........+.AFKPUZ|
| 0x7_ | 64 00 00 2d 37 41 4b 4b 4b 4b 14 03 03 03 03 03  |d..-7AKKKK......|
| 0x8_ | 00 00 2d 32 3c 41 46 4b 59 00 00 32 3c 4b 50 5a  |..-2<AFKY..2<KPZ|
| 0x9_ | 5a 00 0a 02 07 02 02 02 06 19 7d 02 0b 7d 25 00  |Z.........}..}%.|
| 0xa_ | 31 34 46 31 45 4d 53 31 2e 32 30 37 30 32 32 34  |14F1EMS1.2070224|
| 0xb_ | 32 30 32 33 30 39 3a 32 33 3a 34 39 00 00 00 08  |202309:23:49....|
| 0xc_ | 00 00 07 25 00 00 00 00 00 00 00 00 00 00 00 00  |...%............|
| 0xd_ | 00 00 c1 83 0d 00 05 80 00 01 00 00 00 01 00 00  |................|
| 0xe_ | e2 00 00 6b 0e 00 00 00 00 04 00 00 00 84 01 00  |...k............|
[...]

PR #307, opened 2025-05-11
Add support for Prestige 14 Evo B13M and update few devices

mutchiko:
fixes https://github.com/BeardOverflow/msi-ec/issues/306


glpnk:
This does not need new config

PR #365, opened 2025-06-29
Add support for MSI GP76 11UX - E17K3IMS.322

NamitBhutani:


glpnk:
Have you tested this? it will not work

NamitBhutani:
> Have you tested this? it will not work

What do you think is the issue? Will test it soon.

glpnk:
You getting confused with BIOS and EC version. Please test if your code compiles before pushing. You can amend first commit and fix code tabulation.

NamitBhutani:
> ease test if your code compiles before pushing

Sorry, I put the bios version by mistake as they are similar, the code should be correct now.

Also, on further checking the code, I think I can just add `17K3EMS1.115 `to ALLOWED_FW_19 right?

glpnk:
Yeah, ignoring fact that this devices have different screen size. This doesn't matter anyway.

Do you have GP76 Leopard? Right?

Not sure if you have this, and if original device should have it too
```diff
+-			{ SM_SPORT_NAME,   0xc0 },
```

You also need to fix FN-Win swap - it not existed when this device was added, and disable keyboard backlight

```diff
.fn_win_swap = {
		.address = 0xe8,
		.bit     = 4,
-		.invert  = false,
+		.invert  = true,
	},
```
```diff
.kbd_bl = {
-		.bl_mode_address  = MSI_EC_ADDR_UNKNOWN,
+		.bl_mode_address  = MSI_EC_ADDR_UNSUPP,
		.bl_modes         = {},
		.max_mode         = 1,
-		.bl_state_address = 0xd3,
+		.bl_state_address = MSI_EC_ADDR_UNSUPP,
		.state_base_value = 0x80,
		.max_state        = 3,
	},
```

NamitBhutani:
> Yeah, ignoring fact that this devices have different screen size. This doesn't matter anyway.
> 
> Do you have GP76 Leopard? Right?
> 
> Not sure if you have this, and if original device should have it too
> 
> ```diff
> +-			{ SM_SPORT_NAME,   0xc0 },
> ```
> 
> You also need to fix FN-Win swap - it not existed when this device was added, and disable keyboard backlight
> 
> ```diff
> .fn_win_swap = {
> 		.address = 0xe8,
> 		.bit     = 4,
> -		.invert  = false,
> +		.invert  = true,
> 	},
> ```
> 
> ```diff
> .kbd_bl = {
> -		.bl_mode_address  = MSI_EC_ADDR_UNKNOWN,
> +		.bl_mode_address  = MSI_EC_ADDR_UNSUPP,
> 		.bl_modes         = {},
> 		.max_mode         = 1,
> -		.bl_state_address = 0xd3,
> +		.bl_state_address = MSI_EC_ADDR_UNSUPP,
> 		.state_base_value = 0x80,
> 		.max_state        = 3,
> 	},
> ```

I don't have any Shift Mode that sets the value to 0xc0 on 17K3EMS1.115 atleast, can't test it on 1543EMS1.113, in this case would it be better to maintain different configs for each EC firmware version?

glpnk:
keep this shift like in 19 conf, it shouldn't not cause problems

on most devices it's same shift in mean of power settings

glpnk:
have you tested module with your patch? i've squashed all commits into one and fixed something

NamitBhutani:
> have you tested module with your patch? i've squashed all commits into one and fixed something

Yes, tested it MControlCenter, changed settings and hex dumped.

PR #389, opened 2025-07-22
dkms fixes, version bump

glpnk:
dkms.conf: fix CLEAN, dest location, add array indexes
Makefile: change TARGET to KERNELRELEASE from dkms, remove CFLAGS_MODULE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-VERSION         := 0.08
+VERSION         := 0.11
 DKMS_ROOT_PATH  := /usr/src/msi_ec-$(VERSION)
-TARGET ?= $(shell uname -r)
+KERNELRELEASE ?= $(shell uname -r)
 
 ccflags-y := -std=gnu11 -Wno-declaration-after-statement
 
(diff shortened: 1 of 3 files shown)


glpnk:
@SteelT1 Hi, do you have access to AUR package?  https://aur.archlinux.org/packages/msi-ec-dkms-git

SteelT1:
@glpnk Hi, yeah I do, I also maintain that AUR package.

SteelT1:
> @glpnk Hi, yeah I do, I also maintain that AUR package.

Follow up on this on that I have an updated PKGBUILD for that package that now uses the upstream dkms.conf pending for whenever this gets merged into main.

glpnk:
okay, lets merge

glpnk:
but we need to find way to remove old version 

@SteelT1 do you have any ideas?

Because version bumped modules, installed without removing old module creates chaos in dkms tree

SteelT1:
> but we need to find way to remove old version
> 
> @SteelT1 do you have any ideas?
> 
> Because version bumped modules, installed without removing old module creates chaos in dkms tree

I dunno about other distros, but on Arch Linux, the DKMS package provides pacman hooks that removes old versions from dkms tree on package updates

glpnk:
Oh, sounds good, lets try. Reply when you can update AUR package, so I'll merge this

SteelT1:
@glpnk I can update AUR package

glpnk:
Okay, I see we are not able to sync

glpnk:
I will merge now and you update when you can

glpnk:
thanks, @SteelT1

Issue #417, opened 2025-08-08
[Support request] MSI Crosshair 17 B12UGZ 17L3EMS1.106


ios7jbpro:
### Laptop model

[MSI] Crosshair 17 B12UGZ

### EC firmware version

17L3EMS1.106

### EC Dump

```text
|      | _0 _1 _2 _3 _4 _5 _6 _7 _8 _9 _a _b _c _d _e _f
|------+------------------------------------------------
| 0x0_ | 00 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x1_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x2_ | 00 00 00 00 00 00 00 00 0a 05 00 00 00 0c 0b 0b  |................|
| 0x3_ | 03 09 00 0d 00 00 50 81 2b 16 28 3c 37 02 e0 00  |......P.+.(<7...|
| 0x4_ | 00 00 64 00 3a 11 00 00 3a 11 70 40 00 0c 00 00  |..d.:...:.p@....|
| 0x5_ | 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  |................|
| 0x6_ | 00 00 00 00 00 00 00 00 2e 00 37 3e 45 4c 52 58  |..........7>ELRX|
| 0x7_ | 64 28 28 32 37 3c 41 46 50 64 08 04 04 02 03 03  |d((27<AFPd......|
| 0x8_ | 00 00 37 3d 43 48 4d 52 61 00 00 23 28 2d 37 41  |..7=CHMRa..#(-7A|
| 0x9_ | 50 64 08 03 03 03 03 03 06 0f 7d 06 0a 78 2f 00  |Pd........}..x/.|
| 0xa_ | 31 37 4c 33 45 4d 53 31 2e 31 30 36 30 33 32 31  |17L3EMS1.1060321|
| 0xb_ | 32 30 32 32 31 35 3a 31 30 3a 33 35 00 00 00 08  |202215:10:35....|
| 0xc_ | 00 00 01 25 00 00 00 00 00 ca 00 00 00 00 00 00  |...%............|
| 0xd_ | 00 00 c1 80 00 00 05 e4 00 07 00 00 00 03 00 00  |................|
| 0xe_ | e2 00 00 3a 11 01 00 40 00 00 00 00 00 c2 00 00  |...:...@........|
[...]

Issue #347, opened 2025-06-08
MSI Katana A15 AI B8V


EDtheWARDen:
### Laptop model

MSI Katana A15 AI B8VF

### EC firmware version

158NIMS1.505

### List of addresses and their values

```text
EC MSI 158NIMS1.5051204202416:16:41
Type:Embedded Controller: 0066,0062

# Buttons
## [Volume toggle] (Keyboard Light, F1)
2D=00 #Volume unmuted
2D=02 #Volume muted

## [Microphone toggle] (Keyboard Light, F5)
2C=00 #Microphone unmuted
2C=02 #Microphone muted

## [User Scenario] (F7)
### MSI AI Engine (Balanced with AI)
D2=C1
D6=05
### Extreme Performance
D2=C4
D6=03
### Balanced
D2=C1
D6=05
### ECO-Silent
D2=C2
D6=05
(D3=80 #Keyboard light turns off)
(D4=0D #It turned to 0D)

## [Leyboard Light] (F8)
D3={80-83} #User Scenario "MSI AI Engine", "Extreme Performance" and "Balanced" shares mode. ECO-Silent always turns off D3=80. Quirk: There are 5 ticks to the backlight so it rotates on 4 values in Windows 11.

## [Cooler Boost toggle] (Fn+Right arrow, high rpm fan)
98=06 #Cooler Boost off
98=86 #Cooler Boost on

#EC behaviour (ED=xx)
## [ED=B7]
After pressing light down "Fn+F9"
## [ED=B8]
After pressing light up "Fn+F10"
## [ED=C2]
1 second after User Scenario
## [ED=C0]
Changing User Scenario
Pressing Fn
## [ED=C5]
Toggle Microphone
## [ED=C6]
Toggle volume

## [USB Power Share] (MSI Center)
BF=08 #off
BF=28 #on

##[Display OverDrive] (MSI Center)
### Discrete GPU RTX 4060 (laptop)
2E=79 #on, default
2E=69 #off
### Hybrid GPU 780M + RTX 4060 (laptop)
2E=39 #Locked in MSI Control

## [Switch Key] (Switches "Windows/Meta" key with "Fn")
E8=00 #Normal, default
E8=10 #Switch

# [Fan Speed] Values 00-96 = 0%-150%
## Fan 1
71=00 #Default fan speed at cold temp #Changes second
72=00 #Default fan speed at cold temp #Changes first
73=2D #Default at temp 1
74=36 #Default at temp 2
[...]

PR #440, opened 2025-08-20
Add Crosshair 17 B12UGZ

ios7jbpro:
everything matches here. first, compared the values on rweverything, then added the line and built the kernel. features work just fine, no need to add new addresses

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1030,6 +1030,7 @@ static const char *ALLOWED_FW_14[] __initconst = {
 	"17L1EMS1.106", // Katana GF76 11UG
 	"17L1EMS1.107", // Katana GF76 11UE
 	"17L2EMS1.108", // Katana 17 B11UCX, Katana GF76 11UC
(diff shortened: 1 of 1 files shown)


ios7jbpro:
image
image
image
image


just incase if screenshots are required. the issue i opened in the past:
https://github.com/BeardOverflow/msi-ec/issues/417

glpnk:
This will be fastest merge

ios7jbpro:
oh yea and also, from this issue:
https://github.com/BeardOverflow/msi-ec/issues/347
running "echo 'Master Playback Switch' | sudo tee /sys/class/sound/ctl-led/speaker/card2/attach" gets the volume mute led working. just incase if anyone needs it

glpnk:
not a `msi-ec` issue, we don't know who might help, but mute leds are general linux problem

glpnk:
basically, sound card is not mapping it's current out to mute led trigger, but this trigger is successfully bond to driver's led

PR #475, opened 2025-09-20
Add cpu and gpu fan_speed_rpm

Jiogo18:
Add the fan_speed_rpm entry to use the address 0xcc..0xcd (or 0xc8..0xc9) for CPU fan and 0xca..0xcb for GPU fan, like MControlCenter does.
Discussion with @mutchiko https://github.com/dmitry-s93/MControlCenter/pull/267

## realtime_fan_speed

The current `realtime_fan_speed` (0x71 and 0x89) is the fan % target speed for fan mode.
Ideally, it should include "target", "driving" or "control" in its name.

For example, when cooler boost is enabled, fan mode continues to change the target speed according to the temperature.
- dGPU at D3cold, 0x89 is 0, while fan_speed_rpm indicates 6075 rpm.
- For the CPU fan, the temperature decreases and so does the value at 0x71, while fan_speed_rpm reports 8275.

For now, I named the new value fan_speed_rpm to distinguish both.

## About the magic number 480000
- [MControlCenter uses 470000](https://github.com/dmitry-s93/MControlCenter/blob/7bd388b59398b51215baafe2f6c18d12d4a8cd24/src/operate.cpp#L174)
- [ISW uses 478000](https://github.com/YoyPa/isw/blob/7c88670504ecd4462b1825a55bdb0be2944dfe94/isw#L414)
- Last year, I compared with MSI Center and got `floor(480000/x)` (e.g. target speed at 50% ⇒ read 160 ⇒ 3000 rpm)

But now I wonder if MSI Center report is good. RPM above 4000 seems unreal.
By analyzing the audio with my CPU fan, I get
|target speed|Frequency|480000/n|
|-|-|-|
|50%|1980 Hz|3000 rpm|
|100%|3930 Hz|6000 rpm|
|150%|5470 Hz|8275 rpm|

This is more likely 320000 for my CPU fan, with 100% speed targeting 4000 rpm.

https://download-2.msi.com/archive/mnu_exe/mb/MSICENTER.pdf slide 97, graphs show different RPM at 100% (~2100, 2702, and ~4500). It's also not impossible that they have different numerators, it could be a fan_speed_rpm_magic_number.

### DIFF
--- a/ec_memory_configuration.h
+++ b/ec_memory_configuration.h
@@ -58,11 +58,13 @@ struct msi_ec_fan_mode_conf {
 struct msi_ec_cpu_conf {
 	int rt_temp_address;
 	int rt_fan_speed_address; // realtime % RPM
+	int fan_speed_rpm_address_0; // 480000/n RPM
 };
 
 struct msi_ec_gpu_conf {
 	int rt_temp_address;
 	int rt_fan_speed_address; // realtime % RPM
+	int fan_speed_rpm_address_0; // 480000/n RPM
 };
 
 struct msi_ec_led_conf {
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -691,10 +691,12 @@ static struct msi_ec_conf CONF_G1_8 __initdata = {
 	.cpu = {
 		.rt_temp_address      = 0x68,
 		.rt_fan_speed_address = 0x71,
+		.fan_speed_rpm_address_0 = 0xcc,
 	},
 	.gpu = {
 		.rt_temp_address      = 0x80,
 		.rt_fan_speed_address = 0x89,
+		.fan_speed_rpm_address_0 = 0xca,
 	},
 	.leds = {
 		.micmute_led_address = MSI_EC_ADDR_UNSUPP,
@@ -2206,6 +2208,23 @@ static ssize_t cpu_realtime_fan_speed_show(struct device *device,
 	return sysfs_emit(buf, "%i\n", rdata);
 }
 
+static ssize_t cpu_fan_speed_rpm_show(struct device *device,
+					       struct device_attribute *attr,
+					       char *buf)
+{
+	u8 rdata[2];
+	int result;
+
+	result = ec_read_seq(conf.cpu.fan_speed_rpm_address_0, rdata, 2);
+	if (result < 0)
+		return result;
+
+	int value = (rdata[0] << 8) | rdata[1];
+	int rpm = value != 0 ? 480000 / value : 0;
+
+	return sysfs_emit(buf, "%i\n", rpm);
+}
+
 static struct device_attribute dev_attr_cpu_realtime_temperature = {
 	.attr = {
 		.name = "realtime_temperature",
@@ -2222,9 +2241,18 @@ static struct device_attribute dev_attr_cpu_realtime_fan_speed = {
 	.show = cpu_realtime_fan_speed_show,
 };
 
+static struct device_attribute dev_attr_cpu_fan_speed_rpm = {
+	.attr = {
+		.name = "fan_speed_rpm",
+		.mode = 0444,
+	},
+	.show = cpu_fan_speed_rpm_show,
+};
+
 static struct attribute *msi_cpu_attrs[] = {
 	&dev_attr_cpu_realtime_temperature.attr,
 	&dev_attr_cpu_realtime_fan_speed.attr,
+	&dev_attr_cpu_fan_speed_rpm.attr,
 	NULL
 };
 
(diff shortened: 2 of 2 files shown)


glpnk:
You don't mentioned your device, but I will assume it's new enough, like Intel core 11 gen+ or AMD 6 gen+.

We have `msi-wmi-platform` driver which returns fans RPM through `hwmon API`, which can be viewed using `lm-sensors`.

> This is more likely 320000 for my CPU fan, with 100% speed targeting 4000 rpm.

You should not compare different types of fans.

> dGPU at D3cold, 0x89 is 0, while fan_speed_rpm indicates 6075 rpm.

Cooler Boost just bypasses fan curve completely. You should not account it as rule.

> By analyzing the audio with my CPU fan, I get

Do you have broken fan blade, or any other audible marker? Maybe you can get laser tachometer readings as proof?

Jiogo18:
MSI Bravo 15 A4DDR
CPU: AMD Ryzen 5 4600H
GPU: AMD Radeon RX 5500M

I have trouble using sensors. I don't have CPU fan, and I only have GPU fan1 at 65535 RPM.
Does msi-wmi use this exact value? Relying on msi-ec seems more straightforward.

On the other hand, isn't this the same case as for realtime_temperature also reported with sensors?

> Cooler Boost just bypasses fan curve completely

Yes, it illustrates realtime_fan_speed is not the real/current fan speed.

> You should not compare different types of fans.

True, I was confused because I got no frequency spike at the exact rpm. But by looking again, I also get spikes above rpm.
So forget it, I will trust existing values.

glpnk:
Okay, your model is "WMI1", so `msi-wmi-platform` is useless for you.

> I have trouble using sensors. I don't have CPU fan, and I only have GPU fan1 at 65535 RPM.

How many fans you exactly have?

If you have single fan, than it probably follows only CPU fan curve

Jiogo18:
The laptop has 2 fans, one "CPU" and one GPU. GPU fan is often off.

<details>
<summary>sensors</summary>

iGPU, CPU?
```
amdgpu-pci-0800
Adapter: PCI adapter
vddgfx:        1.09 V
vddnb:       937.00 mV
edge:         +43.0°C
PPT:           4.00 W
sclk:         400 MHz
```

GPU off
```
amdgpu-pci-0300
Adapter: PCI adapter
vddgfx:           N/A
ERROR: Can't get value of subfeature fan1_min: Can't read
ERROR: Can't get value of subfeature fan1_max: Can't read
fan1:             N/A  (min =    0 RPM, max =    0 RPM)
edge:             N/A  (crit = +100.0°C, hyst = -273.1°C)
                       (emerg = +105.0°C)
junction:         N/A  (crit = +100.0°C, hyst = -273.1°C)
                       (emerg = +105.0°C)
mem:              N/A  (crit = +105.0°C, hyst = -273.1°C)
                       (emerg = +110.0°C)
ERROR: Can't get value of subfeature power1_cap: Can't read
PPT:              N/A  (cap =   0.00 W)
pwm1:             N/A
sclk:             N/A
mclk:             N/A
```

GPU on
```
amdgpu-pci-0300
Adapter: PCI adapter
vddgfx:        6.00 mV
fan1:        65535 RPM  (min =    0 RPM, max = 3350 RPM)
edge:         +48.0°C  (crit = +100.0°C, hyst = -273.1°C)
                       (emerg = +105.0°C)
junction:     +48.0°C  (crit = +100.0°C, hyst = -273.1°C)
                       (emerg = +105.0°C)
mem:           +0.0°C  (crit = +105.0°C, hyst = -273.1°C)
                       (emerg = +110.0°C)
PPT:           4.00 W  (cap =  85.00 W)
pwm1:             12%
sclk:           0 Hz
mclk:         100 MHz
```
</details>

Jiogo18:
I don't know why I kept this as a draft for so long. It's ready, expect it only contain address for my laptop.

> Okay, your model is "WMI1", so msi-wmi-platform is useless for you.

If I understand correctly from your comments, msi-wmi-platform is for WMI2 and not implemented (yet?) for WMI1?

Is it why `cat /sys/class/hwmon/hwmon5/fan1_input` reports 65535?
temp1_input works well for CPU and GPU (just off by 2-4 degrees with hwmon1 the CPU).
hwmon1/fan1_input doesn't exist, perhaps it's not considered as the CPU fan?

I agree having access from standard API would be great. If you have an idea how I can help, let me know.
Though, I find it more straightforward (for MControlCenter and equivalent) to read directly from msi-ec.

glpnk:
> Is it why cat /sys/class/hwmon/hwmon5/fan1_input reports 65535?

I see this a reason to keep this as draft

Jiogo18:
hwmon:
- GPU: `hwmon5/name` is *amdgpu*. By debugging, found that [drivers/gpu/drm/amd/pm/swsmu/smu11/navi10_ppt.c](https://github.com/torvalds/linux/blob/4a298a43f5e339f48d2dda0665c02b88ee9a4e03/drivers/gpu/drm/amd/pm/swsmu/smu11/navi10_ppt.c#L720) returns `0xffff`. Is it possible that the fan is not connected to the GPU?
- CPU: `hwmon1/name` is *acpitz*. It should contain `fan1_input`, right?

Same result with linux-lts.

The feature is functional and ready. If you prefer not to support it, I'd appreciate your thoughts, so I can focus on troubleshooting hwmon in a discussion instead.

mohammad-abbas-mehdi:
Thanks for this work — this is exactly the gap I've been looking at too. One distinction worth discussing: this PR adds the RPM value as a new platform sysfs attribute, which works but requires tools to know about msi-ec's specific paths. I'm working on an alternative that registers the same data through the hwmon subsystem (`fan1_input` / `fan2_input`), which would make it visible to `sensors`, `fancontrol`, desktop extensions like Vitals, and any hwmon-aware tool without msi-ec-specific knowledge. I'll file a separate issue for that approach — happy to coordinate if useful.

Issue #302, opened 2025-05-10
Support for Vector A18 HX A9WHG


aishen944:
### Laptop model

Vector A18 HX A9WHG

### EC firmware version

 E182LAMS.111(182LIMS1.1050418202510:50:21)

### List of addresses and their values

```text
User Scenario:
Extreme Performance: 0xd2 C4  0xd3 82 
Balanced:0xd2 C1 0xd3 82
ECO-SIent:0xd2 C2 0xd3 80
```

```text
Battery Master:
Best for Mobility: 0xd7 E4
Balanced:0xd7 D0
Best for Battery : 0xd7 BC
```

```text
EC dump:
Type:Embedded Controller: 0066,0062
Width:01
00=00 01=80 02=00 03=00 04=00 05=00 06=00 07=00 
08=00 09=00 0A=00 0B=00 0C=00 0D=00 0E=00 0F=00 
10=00 11=00 12=00 13=00 14=00 15=00 16=00 17=00 
18=00 19=00 1A=00 1B=00 1C=00 1D=00 1E=00 1F=00 
20=00 21=00 22=00 23=00 24=00 25=00 26=00 27=00 
28=0A 29=05 2A=00 2B=00 2C=00 2D=00 2E=3B 2F=5A 
30=03 31=01 32=00 33=0D 34=01 35=00 36=50 37=81 
38=6A 39=18 3A=60 3B=3B 3C=71 3D=02 3E=C0 3F=00 
40=35 41=0C 42=41 43=00 44=5A 45=18 46=00 47=00 
48=D1 49=0F 4A=18 4B=3E 4C=D9 4D=0B 4E=F8 4F=43 
50=00 51=00 52=00 53=00 54=00 55=02 56=00 57=00 
58=00 59=00 5A=00 5B=00 5C=00 5D=00 5E=00 5F=00 
60=00 61=00 62=00 63=00 64=65 65=00 66=B9 67=00 
68=48 69=00 6A=32 6B=38 6C=40 6D=48 6E=50 6F=58 
70=69 71=2D 72=14 73=19 74=1E 75=23 76=2D 77=37 
78=41 79=64 7A=05 7B=03 7C=03 7D=03 7E=03 7F=03 
80=34 81=00 82=32 83=38 84=3E 85=46 86=4B 87=50 
88=64 89=14 8A=00 8B=14 8C=19 8D=23 8E=2D 8F=37 
[...]

PR #556, opened 2025-12-09
feat: Add MSI Raider A18 HX A7VIG 182KIMS1

monokrome:
HI!

In case you're interested, this is what I use on my Raider A18 HX A7VIG. It shares a chassis with the other ones here. The fan controls, shift mode, and cooler boost all work properly and temperature readouts seem to be accurate.

One thing that's worth noting is that the webcam controls do not work. I'm guessing this model doesn't have a way to do a software disable on the webcam since it has a hardware shutter, but its also possible that I just didn't find the address for the webcam stuff in my probing.

I figured I'd still offer the patch since webcam seems non-essential when the device otherwise crashes without proper fan support. I've been using this for about a month now and it has prevented the device from crashing. The device crashes a lot during more intensive work if this isn't enabled.

Thanks for the great project!

### DIFF
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -1637,6 +1637,7 @@ static const char *ALLOWED_FW_G2_10[] __initconst = {
 	"1824EMS1.107", // Titan 18 HX Dragon Edition
 	"182LIMS1.108", // Vector A18 HX A9WHG
 	"182LIMS1.111", // New ec version for Vector A18 HX A9WHG
+	"182KIMS1.113", // Raider A18 HX A7VIG
 	NULL
 };
 


glpnk:
What about charge control? Because 7 gen Intel CPU based models were mostly unable to control bat limit. This feature was added around 8 gen.

monokrome:
It looks like charge control isn't working fully. I can read battery level but the thresholds read out as 0. This is an AMD CPU, no intel here. Something I can look into.

glpnk:
I can't help you without EC dump. Please open issue and answer all questions
https://github.com/BeardOverflow/msi-ec/issues/new?template=support_request.yml

monokrome:
I didn't ask for help? I said I'd look into it. Also, pretty sure its working - I just had to set a charge control threshold and it reads back as expected. I'll keep testing to verify.

    polar@sirius % echo 80 | sudo tee /sys/class/power_supply/BAT1/charge_control_end_threshold
    80
    polar@sirius % cat /sys/class/power_supply/BAT1/charge_control_end_threshold
    80

monokrome:
Charge control is working.

If you'd like more details on how I've tested, I wrote 80 to the end threshold. I pulled my AC cable out, waited until it was at 77% (with end threshold at 80% and start threshold at 70%) and plugged back in. Got this:

    polar@sirius % cat /sys/class/power_supply/BAT1/capacity /sys/class/power_supply/BAT1/status /sys/class/power_supply/ADP1/online
    77
    Not charging
    1

Unplugged again, waited until sub-70, got this:

    polar@sirius % cat /sys/class/power_supply/BAT1/capacity /sys/class/power_supply/BAT1/status /sys/class/power_supply/ADP1/online
    69
    Charging
    1

It maintains charging after hitting start threshold:

    polar@sirius % cat /sys/class/power_supply/BAT1/capacity /sys/class/power_supply/BAT1/status /sys/class/power_supply/ADP1/online
    71
    Charging
    1

It should stop charging at 80 since msi-ec doesn't store the start/end separately and just subtracts 10 from the end instead. I can update when its done charging.

monokrome:
Stopped charging at 80. Charge control works, fans work. Only thing I'm aware of that isn't working is the webcam:

    bat -p /sys/class/power_supply/BAT1/capacity /sys/class/power_supply/BAT1/status /sys/class/power_supply/ADP1/online
    80
    Not charging
    1

glpnk:
Hi, your laptop looks similar to https://github.com/BeardOverflow/msi-ec/issues/302, but it supports webcam control. 

> https://github.com/BeardOverflow/msi-ec/issues/new?template=support_request.yml

Please complete survey, it will be documentation about your device for other owners. EC dump will help to figure situation with webcam control. It might be possible that hardware reports that supports control, but IRL - not support it. I hope you've not wrote to `webcam_block` file.

Thanks for your contribution.

glpnk:
More about webcam - some old devices had hardware switch, state of which was reported to `webcam_block` file. But your device is second gen, which had no such switches. If you have only cover and no button on keyboard - you probably have no control on webcam power.

PR #595, opened 2026-01-22
Add support for MSI GE76 Raider 11UE (17K3EMS1.113)

Diamon0:
It worked on a previous version, but after a certain update, it stopped working, so I took it upon myself to make it work again. It's working same as before it broke on my laptop, and I can't see anything going wrong after a full day of use, kernel updates, and some shut-downs and reboots.


glpnk:
Seems you have downgrade, I don't know how this possible. But I'm adding it

PR #598, opened 2026-01-25
Update which dkms command to include sudo

tsubery:
On debian 13 the binary isn't available for unprivileged users.


glpnk:
Does `dkms status` available in Debian without sudo/root?

tsubery:
Now both `dkms status` and `which dkms` work without sudo on my debian 13 installation. Maybe I accidentally used the wrong terminal the last time.

glpnk:
Okay, maybe you've updated PATH for you user and added one from `root`

PR #613, opened 2026-02-16
Adding new version for Vector A18 HX A9WHG

vbresan:


glpnk:
This does not work this way, you've added BIOS version instead of EC

vbresan:
@glpnk So where can I find EC?

glpnk:
@vbresan 

https://github.com/BeardOverflow/msi-ec/issues/302#issuecomment-3904787688

https://github.com/BeardOverflow/msi-ec/issues/302#issuecomment-3904789407

I've re-formatted your dump from RW everything format to more easily readable text hex and found that you have `182LIMS1.111`

glpnk:
Let's figure what can cause problems:

1. Do you have Secure Boot enabled?
2. Did you used `dkms` to install?
3. Can you send dump of `modinfo msi-ec` after you installed module

PR #639, opened 2026-03-08
Implement USB Powershare feature

Fede2782:
Tested on C1MG

old pr #566

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,13 @@ This driver exports a few files in its own platform device, msi-ec, and is avail
     - on: cooler boost function is enabled
     - off: cooler boost function is disabled
 
+- `/sys/devices/platform/msi-ec/usb_powershare`
+  - Description: This entry allows enabling the usb powershare function. It provides power to charge devices via USB ports while device is suspended.
+  - Access: Read, Write
+  - Valid values:
+    - on: usb powershare function is enabled
+    - off: usb powershare function is disabled
+
 - `/sys/devices/platform/msi-ec/available_shift_modes`
   - Description: This entry reports all supported shift modes.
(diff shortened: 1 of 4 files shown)


glpnk:
why you closed previous PR?

Fede2782:
> why you closed previous PR?

I changed branch named to something which has a actual meaning, also rebased the change to use to the correct values for the G2_0 Fw group which was wrongly placed in old PR.

Also want to make my local clone more organized since I'm currently working on other features (KBD backlight mode change) for this project to complete support for my EC Control app.

Fede2782:
it would be cool if other owners could add their EC adresses if they support this function.

glpnk:
https://codeberg.org/Sparronator9999/YAMDCC/issues/89 there is some leaked doc about EC structure

Fede2782:
> https://codeberg.org/Sparronator9999/YAMDCC/issues/89 there is some leaked doc about EC structure

Cool. Probably all gen2 have this set in 0xbf but I'd like to hear some gen1 user first.

glpnk:
You can read "support bit" `0xBF bit 3` to detect if feature available to dynamically unlock it

But gen 1 seems to be unsupported completely

Fede2782:
> You can read "support bit" `0xBF bit 3` to detect if feature available to dynamically unlock it
> 
> But gen 1 seems to be unsupported completely

Sorry for late reply. That's a great idea but I'm gonna keep it this way so that it matches the rest of the application. In case you believe the bit check is required for you to consider merging this, I'd be more than happy to accordingly overhaul this.

glpnk:
Please migrate to bits

PR #648, opened 2026-03-17
Implement the ability to change the backlight mode

Fede2782:
* The config had values but they were unused up to this point

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,19 @@ Led subsystem allows us to control the leds on the laptop including the keyboard
     - 2: Half
     - 3: Full
 
+- `/sys/class/leds/msiacpi::kbd_backlight/available_kbd_bl_modes`
+  - Description: This entry reports all supported keyboard backlight modes.
+  - Access: Read
+  - Valid values: Newline separated list of strings.
+
+- `/sys/class/leds/msiacpi::kbd_backlight/kbd_bl_mode`
+  - Description: This entry allows switching the keyboard backlight mode. It allows control on when to use the keyboard backlight.
+  - Access: Read, Write
+  - Valid values: 
+    - Values reported by `/sys/class/leds/msiacpi::kbd_backlight/available_kbd_bl_modes`. Some of the possible values:
+      - always: always keep the keyboard backlight enabled.
+      - timeout: automatically disable the keyboard backlight after 10s from last interaction with they keyboard or touchpad.
+
 ### Debug mode
 
 You can use module *parameters* to get direct read-write access to the EC or force-load a configuration
--- a/ec_memory_configuration.h
+++ b/ec_memory_configuration.h
@@ -74,7 +74,7 @@ struct msi_ec_led_conf {
 #define MSI_EC_KBD_BL_STATE_MASK 0x3
 struct msi_ec_kbd_bl_conf {
 	int bl_mode_address;
-	int bl_modes[2];
+	struct msi_ec_mode bl_modes[5];
 	int max_mode;
 
 	int bl_state_address;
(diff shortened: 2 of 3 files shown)


glpnk:
Does keyboard backlight being reliably triggered at all? I'm assuming it triggers only by keys, not by touchpad?

I'm comparing this with how this made by Asus

Fede2782:
> Does keyboard backlight being reliably triggered at all? I'm assuming it triggers only by keys, not by touchpad?
> 
> I'm comparing this with how this made by Asus

here on modern 15 h ai c1mg-096, using touchpad does trigger backlight as well. i set this as WIP cause i'd like someone without the backlight to test it as well.

glpnk:
That's really nice, because my other laptops backlight from some chinese brand does not triggered by touchpad

glpnk:
Does toggling this setting affects backlight? Like idea of making KB glow when mice is moved or screen is touched

Fede2782:
> Does toggling this setting affects backlight? Like idea of making KB glow when mice is moved or screen is touched

This setting only changes the mode. If backlight is on or off it's decided in the brightness attribute. This attribute might be managed by the DE so that keyboard could be triggered using external inputs such as mouse or keyboard. 

When brightness attribute is greater than zero then the brightness mode is followed. If it's "timeout" then every time a key or touchpad is used it will stay on for 10s. If it's "always" then it will always stay on until brightness is set to 0.

glpnk:
You right but system does not know real state of backlight. You can read it, but DE will not update state without a trigger.

Is BL level is 0 when state is timeout and blank?

Fede2782:
> You right but system does not know real state of backlight. You can read it, but DE will not update state without a trigger.
> 
> Is BL level is 0 when state is timeout and blank?

Once this module is installed everything knows and can control the backlight. KDE, for example, recognizes it and allows control of brightness intensity directly via its control panel. Therefore you could write an app to trigger the backlight anytime with any trigger you want since it is available in sysfs for apps.

BL level doesn't change if mode is timeout but the timeout has reached 10s and powered off. It remains untouched cause the keyboard backlight is already on but it has been temporarily disabled because timeout mode is set

glpnk:
I mean that KDE and other DEs can't know that you changed brightness with keyboard key

Fede2782:
> I mean that KDE and other DEs can't know that you changed brightness with keyboard key

They could tho. Once you press the keyboard key the EC value gets updated and therefore the sysfs attribute does as well when being read. So using a watcher they could detect those changes.

glpnk:
Gnome have no watcher built in

Fede2782:
> Gnome have no watcher built in

I never said it has. I am saying that once this module is installed there are a lot of new possibilities and DEs could take advantage of them in every way you could possibly think of.

glpnk:
If you know how to tell DE that brightness is changed, can you say how to do this? At leas on KDE if you know.

This not related to this PR and to this repo at all. But there a design possibility to receive events about brightness and other soft keys through WMI. But this is related to `msi-wmi-platform` driver, not `msi-ec`

Fede2782:
> If you know how to tell DE that brightness is changed, can you say how to do this? At leas on KDE if you know.
> 
> This not related to this PR and to this repo at all. But there a design possibility to receive events about brightness and other soft keys through WMI. But this is related to `msi-wmi-platform` driver, not `msi-ec`

I haven't thought of a precise solution but, as I stated in previous replies, a possible solution could be implementing a watcher which periodically checks the brightness value exported by this driver and does actions depending on that. I didn't consider getting events WMI, could be possible that as well

glpnk:
I think this events are standard because DE can rotate screen based on gyroscope for example, so we need to just tell DE that something happened to update its internal state

PR #649, opened 2026-03-17
Dkms updating target and improved dkms removal script

Xabi08YT:
- [x] Removal of dkms module regardless of the version
- [x] Update target working (to test)

### DIFF
--- a/universalDKMSModuleUninstaller.sh
+++ b/universalDKMSModuleUninstaller.sh
@@ -2,7 +2,7 @@
 
 FILTER="msi-ec|msi_ec"
 
-sudo dkms status | grep -E "$FILTER" | while read -r line; do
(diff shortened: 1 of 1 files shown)


glpnk:
Please remove `sudo` from script

Xabi08YT:
> Please remove `sudo` from script

It is necessary for dkms isn't it ? 
I mean, I prefer having the script elevating some of the commands inside rather than elevating the whole script. But if you really want me to remove sudo, i can.

glpnk:
All except `make`/`make modules` and `make clean` in this script requires root, so `sudo` is redundant.

Thanks

Xabi08YT:
So, does the script seem to be ok for you ?

glpnk:
It works

PR #632, opened 2026-03-05
Akmod packaging

Xabi08YT:
SRPM packaging script for MSI-EC.
Designed to work on Fedora and Fedora Silverblue. Builds may be opened to more RPM-based distro in the future.

Distros:
 - Fedora 42 (Testing not Planned as EOL is in April 2026, community maybe ?, BORKED (RPM Fusion repos OUT))
 - [x] Fedora 43 Silverblue/Kinoite (Installed flawlessly)
 - [x] Fedora 43 Workstation (Installed flawlessly.)
 - Fedora 44 (Not out yet, but build is already opened in case)
 - Fedora Rawhide (Opened to have insights)
 - EPEL 9 (Opened in case)
 - [x] EPEL 10 (Nightly only atm)

Anyway, I stated into the COPR install instruction, that the recommended way for Workstation editions and RHEL remains this github repo and DKMS. COPR are only provided as backup for distros like Fedora Silverblue/Kinoite that are only supporting Akmod.

PR #655, opened 2026-03-20
fix dkms

glpnk:
```
git clone -b dkms-hotfix https://github.com/glpnk/msi-ec.git
cd msi-ec
sudo make dkms-uninstall
sudo make dkms-install
```

fix bug introduced in https://github.com/BeardOverflow/msi-ec/pull/632

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,10 @@ obj-m += $(MODNAME).o
 all: modules
 
 modules:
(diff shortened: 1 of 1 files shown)


glpnk:
@Xabi08YT Why you replaced `CURDIR` with `PWD`? It's empty when run with `sudo`

Xabi08YT:
> @Xabi08YT Why you replaced `CURDIR` with `PWD`? It's empty when run with `sudo`

Replaced it because of akmod. If I was using PWD it would crash everything.

PR #656, opened 2026-03-20
Hotfix akmods

Xabi08YT:


Xabi08YT:
There you go @glpnk 
Here you should have a version that works for akmods and dkms.

glpnk:
It does not crashed last 2 builds I've pushed

Xabi08YT:
It does not break the build, but the install on the client's side

glpnk:
Also why you replaced them back to your state, which don't work with `sudo`?

Xabi08YT:
> Also why you replaced them back to your state, which don't work with `sudo`?

What ? I put back pwd like you did in one of your previous pr but the changes got rolled back look at main

glpnk:
Before akmod PR it used CURDIR, so I've rolled back to it. PWD is empty when executed with sudo

Xabi08YT:
> Before akmod PR it used CURDIR, so I've rolled back to it. PWD is empty when executed with sudo

Wait. Hold on a second.
I thought it was PWD. Can you look at the diffs ?

glpnk:
CURDIR was first introduced in https://github.com/BeardOverflow/msi-ec/commit/5951c6a9fc961059a27bddb61597d89e32aadb64

You switched it to PWD in first commit of akmod PR https://github.com/BeardOverflow/msi-ec/commit/a8e9a28da7ab53768bff1963630bf6ad2ac4ebb3

And I've switched back to CURDIR https://github.com/BeardOverflow/msi-ec/commit/99546400ab54ee93e5ba3b2e266d01c44f39952b

glpnk:
So your sed command changes PWD to CURDIR, but only for akmod. But both worked well for akmod build, but only CURDIR worked for DKMS

Xabi08YT:
> So your sed command changes PWD to CURDIR, but only for akmod. But both worked well for akmod build, but only CURDIR worked for DKMS

Ok, let me change back my sed.
And also, the build will pass no matter if CURDIR or PWD is used. BUT, with CURDIR, if I remember well, it won't install on clients.

Xabi08YT:
Let me test the last akmod real quick. I will keep you updated

glpnk:
Command where you passed PWD is only required for build. Build will work identically well with PWD or CURDIR without sudo

https://github.com/BeardOverflow/msi-ec/blob/2af3c3e0921548142cb080731a91bac2d42bf0a0/Makefile#L16-L21

Plus this only affects build, not install. Not sure about client side build, but should work because CURDIR is internal make command, unless Fedora builder resets CURDIR, obviously

Xabi08YT:
No it affects install because akmods is compiling the module on the fly.

glpnk:
Okay, can you rollback only build and clean options and keep dkms?

Xabi08YT:
> Okay, can you rollback only build and clean options and keep dkms?

No worries, I will just invert the SED so the Makefile stays with curdir for normal installs but switches to PWD for akmods.

glpnk:
Why in COPR's readme mentioning `kmod-msi-ec` for Silverblue, `akmod-msi-ec` for Desktop, but package named `msi-ec-kmod`?

Xabi08YT:
> Why in COPR's readme mentioning `kmod-msi-ec` for Silverblue, `akmod-msi-ec` for Desktop, but package named `msi-ec-kmod`?

Well it is a bit complex. By default, the spec file I sent to this repo is for the kmod. That is why the package is named msi-ec-kmod, as required by kmodtool. However, kmodtool also generates and wants a package msi-ec-kmod-common and akmod-msi-ec, which have specific use cases.

In general, it is better to use the akmod package, because akmods works a bit like dkms, where it builds for every kernel and every new kernel. So it is better to use it. But for silverblue, akmods cannot work properly because these distro are using rpm-ostree. That is why we are using the kmod package instead, which will install the package in a different manner suitable for rpm-ostree.

glpnk:
Done

glpnk:
Can you check that everything works now?

Xabi08YT:
> Can you check that everything works now?

Not now but during the day I will. I forgot to plug back the lan cable on my machine that tests for Fedora 43 Workstation and Rocky Linux 10. For Fedora Silverblue, it will take some minutes because I need to finish a task first.

glpnk:
No problem

Xabi08YT:
I also opened a discussion about this, asking for some people to test if everything works

glpnk:
Can you add link to it into readme? I've pinned it but people rarely check pinned items

Xabi08YT:
Sorry I did not see the message. I am going to do it right now while waiting for the train (late again)

Xabi08YT:
Okay, little update here: The package is working on Rocky Linux 10.1 (just tested). Fedora Silverblue seems to have worked too (I still have the module after the update so...). Fedora workstation tests are delayed until i get physical access to the machine on Sunday evening.

Upgrade process worked flawlessly between the two version that followed the new versioning system with the commit slug.

glpnk:
> Can you add link to it into readme? I've pinned it but people rarely check pinned items

I meant Fedora section

glpnk:
> Okay, little update here: The package is working on Rocky Linux 10.1 (just tested). Fedora Silverblue seems to have worked too (I still have the module after the update so...). Fedora workstation tests are delayed until i get physical access to the machine on Sunday evening.
> 
> Upgrade process worked flawlessly between the two version that followed the new versioning system with the commit slug.

Nice

Xabi08YT:
> > Can you add link to it into readme? I've pinned it but people rarely check pinned items
> 
> I meant Fedora section

Yeah but since the first message of the discussion also asks about dkms, I thought it could be in the contribute section

Xabi08YT:
Would you agree moving this conversation on #658 ?

PR #730, opened 2026-05-24
AI Policy

Xabi08YT:
Added an AI Policy section.

Everything in the AI Policy should still be discussed  as it is important. Feel free to request changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ When making an issue, you must follow the following rules:
 3. Ticket ENTIRELY GENERATED with AI will NOT BE TREATED if the issue is not deemed serious enough
 4. Fully generated AI bug reports may be straight up closed and denied by the maintainer of the project
 5. You MUST be able to provide details when asked for it.
-6. You MUST tell when the AI was involved in the process. For diagnosis, please detail the model used and the agent tool if used in the **FIRST** message of the issue.
+6. You MUST tell when the AI was involved in the process. For diagnosis, please detail the model used and the agent tool if used in the **FIRST** message of the issue or somewhere or in a dedicated field of the issue template.
 
 In general, AI bug reports are discouraged unless it is to help with formulated the content in english because it is not your native language.
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ You can also contribute by sharing your feedback on the installation process and
 
 ### AI Policy for code contributions, documentations contribution
 When contributing code or documentation you have to follow the following rules:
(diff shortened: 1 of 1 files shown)


teackot (approved):
Looks good, seems consistent with how various other FOSS projects try to regulate it

README.md
```diff
@@ -25,6 +25,26 @@ Eager to support the project? Your help is always welcome to keep the project al
 
 You can also contribute by sharing your feedback on the installation process and problems encountered with it in this discussion: https://github.com/BeardOverflow/msi-ec/discussions/662
 
+## AI Policy
+
+### AI Policy for code contributions, documentations contribution
+When contributing code or documentation you have to follow the following rules:
+1. Explicitely tell what has been made by AI and by which LLM
```
glpnk: I think we also need to ask about agent tool if one is used https://docs.kernel.org/process/coding-assistants.html#attribution

glpnk:
For other things looks good

Xabi08YT:
I just changed the AI policy to include your feedback. I think we should also include some fields on the Issue template for AI use.

glpnk:
Maybe to Bug template, like used tools. But not to request, or check mark at maximum.

Xabi08YT:
Yeah true. This is something that also needs to be discussed.

README.md
```diff
@@ -25,6 +25,27 @@ Eager to support the project? Your help is always welcome to keep the project al
...
+In general, AI code is strongly discouraged and may get more negative reviews by the maintainers.
+
+### AI Policy for issues
+When making an issue, you must follow the following rules:
+1. You may use AI to help get some infromations and commands to enrich your ticket
+2. You may use AI to improve the grammar and readability of your ticket
+3. Ticket ENTIRELY GENERATED with AI will NOT BE TREATED if the issue is not deemed serious enough
+4. Fully generated AI bug reports may be straight up closed and denied by the maintainer of the project
+5. You MUST be able to provide details when asked for it.
+6. You MUST tell when the AI was involved in the process. For diagnosis, please detail the model used and the agent tool if used in the **FIRST** message of the issue.
```
glpnk: Let's ask tool names in `Additional info` or `Description` section.  Because there is no first message in templates.

glpnk: Having 2 different names for same free form field is confusing and unnecessary. Will be good to rename one of them

glpnk:
Interestingly that `gh cli` allowed (or still allow) to create free form Issues in the past. There were 1 free form issue made this way. Agents might use it to send info, but it's unknown for most users

PR #773, opened 2026-07-11
Drop minor version (NEEDS TESTING)

### LINKED ISSUE #32: Add fallback mode
As far as I can tell, EC FW versions that only differ in the last two digits are backwards compatible. We may want to add an optional fallback mode, so users with newer EC FW versions can still use the driver by falling back to older compatible versions.

We also know that cooler boost address _may be_ universal for all FWs, so we can make a "global fallback". This way users with unsupported firmware will at least be able to use cooler boost.

Xabi08YT:
Hi,

I just finished the rework on version check during module load.
However, I cannot test that since I only have access to WSL atm (Intel please publish the freacking patch).

So if someone could test it would be great. Some regression may occur since I did some deduplicating.

I also tried adding a blacklist but it should also be tested. I know that it is not needed now but it may be used one day or another.

Like always, no AI. So if you want to check.

### DIFF
--- a/Makefile.vars
+++ b/Makefile.vars
@@ -1,2 +1,2 @@
 MODNAME := msi-ec
-VERSION := 0.13
+VERSION := 0.20
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -3020,8 +3020,8 @@ static int __init load_configuration(void)
 	}
 
 	// Truncating the version to only keep what comes before the dot. EG: 15M3EMS1.113 -> 15M3EMS1
-	char ver_truncated[MSI_EC_FW_VERSION_LENGTH - 3];
-	memcpy(ver_truncated, ver, MSI_EC_FW_VERSION_LENGTH - 4);
+	char ver_truncated[MSI_EC_FW_VERSION_LENGTH - 1];
+	memcpy(ver_truncated, ver, MSI_EC_FW_VERSION_LENGTH - 2);
 	ver_truncated[8] = '\0';
 
 	int banned = 0;
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -3128,6 +3128,8 @@ MODULE_LICENSE("GPL");
 MODULE_AUTHOR("Jose Angel Pastrana <japp0005@red.ujaen.es>");
 MODULE_AUTHOR("Aakash Singh <mail@singhaakash.dev>");
 MODULE_AUTHOR("Nikita Kravets <teackot@gmail.com>");
+MODULE_AUTHOR("Xabi Goity <xabigoity@gmail.com>");
+MODULE_AUTHOR("Glitch Punk <glpnk@proton.me>");
 MODULE_DESCRIPTION("MSI Embedded Controller");
 MODULE_VERSION("0.13");
 
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -54,6 +54,12 @@ static DEFINE_MUTEX(ec_set_bit_mutex);
 #define FM_BASIC_NAME		"basic"
 #define FM_ADVANCED_NAME	"advanced"
 
+//Used to ban specific firmware versions in case they are buggy.
+static const char *BANNED_FW_VERSION[] __initconst = {
+	NULL
+};
+
+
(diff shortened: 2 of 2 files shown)


EricLin0509:
@Xabi08YT 

Hi, I just tested your commits, but it cause 2 compilation errors

```
msi-ec.c:60:2: error: expected ';' after top level declarator
   60 | }
      |  ^
      |  ;


msi-ec.c:2831:53: error: expected ';' after expression
 2831 |                 pr_err("Your firmware version is blacklisted !\n")
      |                                                                   ^
      | 
```

I'm using my own [forks](https://github.com/EricLin0509/msi-ec/tree/MinorVersionTest) with your commits

Xabi08YT:
> @Xabi08YT 
> 
> Hi, I just tested your commits, but it cause 2 compilation errors
> 
> ```
> msi-ec.c:60:2: error: expected ';' after top level declarator
>    60 | }
>       |  ^
>       |  ;
> 
> 
> msi-ec.c:2831:53: error: expected ';' after expression
>  2831 |                 pr_err("Your firmware version is blacklisted !\n")
>       |                                                                   ^
>       | 
> ```
> 
> I'm using my own [forks](https://github.com/EricLin0509/msi-ec/tree/MinorVersionTest) with your commits

Thank you very much. I will correct that right away

Xabi08YT:
@EricLin0509 Thank you it should be fixed

Xabi08YT:
Hello @glpnk,
I just wanted to check if you had time to test the PR ?
Also, small question, do we keep the version check like this or do we edit it to use the sku number as you suggested ?

glpnk:
Sorry for the slow reply, but I think it is worth dropping only the last 2 digits of the version, not all 3 with + dot.

The SKU is the BIOS version, which also, probably printed on the board as MS-xxxxy, but no proofs. At the current stage of config list shrink, we are only dropping the version (or release number) part.

In the case of EC versions and features, we currently have 6 viable bytes: `xxxx_MSy.z--`. In place of the `_` is the EC chip brand ITE/ENE. In place of `MS`, there can be different values, like `WS` for workstation, and few pre-release codes, which are questionable to support. Basically, we have no samples of those devices.

Xabi08YT:
Ok I see. I will correct that later today

Xabi08YT:
But you know, since the .1 duplicate data and almost never changes I thought dropping it would not change much

Xabi08YT:
Yeah ok i just saw a .2 so yes we should keep it then. I am resolving the merge conflict so I will redo most of the deduplicating

Xabi08YT:
@glpnk I corrected like you said. However it needs to be tested still. Some regression may occur but I think it should be fine

glpnk:
Sorted and squashed commits

[Backup branch](https://github.com/glpnk/msi-ec/tree/DropMinorVersionBackup)

<details><summary>📩</summary>At least it compiles</details>

Xabi08YT:
OK wow good job @glpnk 
Much cleaner than what I did lol. But i kinda hate making regex lol

glpnk:
Well, it was joke. Anyway I've deduped them manually

Xabi08YT:
> Well, it was joke. Anyway I've deduped them manually

Wow thx. It is really a painful job

Xabi08YT:
OK I see what you meant now. Yeah right it is also good.

Xabi08YT:
I think everything looks good. You can merge if you want

Xabi08YT:
Hey guys, just wanted to know if there is any conflicts that need to be resolved ATM before merging this.

glpnk:
I'll fix them and merge soon

glpnk:
I think we can drop bigger part of the name, because recently we got request from Workstation/Creator devices and devices made for Turkey, but maybe address this problem in newer PR after this

Xabi08YT:
OK. But for dropping bigger part of the name idk. I don't feel like it is a good idea. I think that would make that too broad. IDK.

Xabi08YT:
So maybe we should start like this, and drop more later if we feel like it is something needed

Xabi08YT:
Hey @glpnk I just updated this branch to match main.

PR #788, opened 2026-08-01
Remove Super Battery

glpnk:
What was assumed as Super Battery actually been CPU Pstate


Xabi08YT:
Small question:

What would that actually change ? I mean, if you can disable pstate then on battery it is a good thing lol

glpnk:
It's unused, and Pstates are set by shift mode. Adding this setting been wrong from first releases

Xabi08YT:
Oh yeah then useless.

PR #798, opened 2026-08-07
Add support for rear 'STEALTH' leds on 14K1EMS1 and 14K2EMS1 models

bluer222:
Adds support for the leds on the rear of MSI stealth laptops for `14K1EMS1.108` (Stealth 14 Studio A13V)

**Changes:**
- add stealth_led_address and stealth_led_bit (as it uses a different bit than the existing keyboard leds)
- add logic for handling stealth led (exactly the same as mic and micmute leds)
- create `CONF_G2_11`, a direct copy of `CONF_G2_5`
- set `14K1EMS1.108` to use the new `CONF_G2_11`
- set `stealth_led_address = 0x2c` and `stealth_led_bit = 4` in `CONF_G2_11`

**Testing:**
- windows method used to find address and value for toggling the led
- Verified with RWEverything and msi-ec debug: writing 0x2c = 0x10 turns on the stealth LED; 0x2c = 0x12 turns on stealth+micmute; 0x2c = 0x00 turns both off.
- Built and loaded the kernel module with these changes; toggling `/sys/class/leds/platform::stealth/brightness` between 1 and 0 turns the rear light on and off without any other effects.

**Note:**
- setting `/sys/devices/platform/msi-ec/shift_mode` to `eco` turns off the stealth led regardless of the value set in `/sys/class/leds/platform::stealth/brightness`, this is the same behavior as using the official MSI Center app

### DIFF
--- a/ec_memory_configuration.h
+++ b/ec_memory_configuration.h
@@ -69,6 +69,8 @@ struct msi_ec_led_conf {
 	int micmute_led_address;
 	int mute_led_address;
 	int bit;
+	int stealth_led_address;
+	int stealth_led_bit;
 };
 
 #define MSI_EC_KBD_BL_STATE_MASK 0x3
--- a/msi-ec.c
+++ b/msi-ec.c
@@ -117,6 +117,8 @@ static struct msi_ec_conf CONF_G1_0 __initdata = {
 		.micmute_led_address = 0x2b,
 		.mute_led_address    = 0x2c,
 		.bit                 = 2,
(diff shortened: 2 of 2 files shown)


kyokenn:
I can confirm that stealth leds work on 14K2EMS1 too (Stealth 14 AI Studio A1VFG).
The address and bit is the same.
My commit for reference - https://github.com/kyokenn/msi-ec/commit/3d05a6b5d020974642bd0e520f05877b2437d262

PR #806, opened 2026-08-14
Add support for EC firmware 15P2ED61.100 

HIbrahimK:
Add support for EC firmware 15P2ED61.100 GameGaraj Slayer 4 (MSI Sword 16 HX derivative)


glpnk:
I'm curious, can you flash official firmware and turn GameGaraj into MSI? `ED6` part of name looks like firmware for pre release or prototype models. I'm not asking to risk, but maybe someone tried?

HIbrahimK:
To be honest, it is theoretically possible to force-flash the official MSI Sword 16 HX retail firmware. I have even seen people online doing this using low-level deployment tools like AFUWIN or AFUDOS to completely bypass the subsystem ID checks, similar to the process shown in this video (https://www.youtube.com/watch?v=qWf7QgW4EeI).
However, it comes with a huge risk of permanently bricking the laptop, and you often end up with weird hardware bugs like broken keyboard shortcuts or broken display brightness controls.
Since I am running CachyOS, the Windows-only MSI Center isn't an option for me anyway. On Linux, we rely on the msi-ec kernel module and mcontrolcenter for fan and power profiles.
Actually, I have already solved the issue on my own machine, and I submitted this Pull Request purely to help the rest of the open-source community. By upstreaming support for this specific 15P2ED61.100 EC firmware, other GameGaraj Slayer 4 users won't have to risk bricking their devices with dangerous BIOS flashes just to get native fan control working on Linux. My laptop is also still under active warranty, so I personally prefer keeping it completely stock. Hopefully, this gets merged soon to help out others!

glpnk:
Thanks.

But still weird that MSI is using pre release firmwares on these devices, because they sometimes release updates with bug fixed

